PR #16, opened 2023-09-09
VB-2464 Replace HMPPS branding with default GOV.UK

tpmcgowan:
* Remove MoJ frontend library and associated config
* Remove unnecessary code (IE8 styles, HTML5Shiv, etc)
* Remove jQuery - was required by MoJ Frontend but not by GDS
* Use default GOV.UK page template


integration_tests/e2e/login.cy.ts
```diff
@@ -20,20 +20,20 @@ context('SignIn', () => {
     Page.verifyOnPage(AuthSignInPage)
   })
 
-  it('User name visible in header', () => {
+  it.skip('User name visible in header', () => {
```
tpmcgowan, line 23: Just skipping these tests that relate to HMPPS Auth integration for now - will all be stripped out when moving to GOV.UK One Login...

server/middleware/setUpWebSecurity.ts
```diff
@@ -28,6 +28,7 @@ export default function setUpWebSecurity(): Router {
           styleSrc: ["'self'", (_req: Request, res: Response) => `'nonce-${res.locals.cspNonce}'`],
           fontSrc: ["'self'"],
           formAction: [`'self' ${config.apis.hmppsAuth.externalUrl}`],
+          upgradeInsecureRequests: process.env.NODE_ENV === 'development' ? null : [],
```
tpmcgowan, line 31: Get it to work in Safari on localhost (I should suggest this for the template app...🤔)

PR #20, opened 2023-09-14
VB-2554 Add test API call

tpmcgowan:
Add support for calling the Orchestration Service API and make a test call (to list supported prison IDs).

Code and config, including for tests, etc brought over from Staff and Admin UI apps.

Purpose is to demonstrate that calling APIs with credentials is still working given HMPPS Auth login replaced with GOV.UK One Login - and that testing set up still works. Will help inform what lingering auth code can be removed.

### DIFF
--- a/server/services/supportedPrisonsService.ts
+++ b/server/services/supportedPrisonsService.ts
@@ -1,0 +1,14 @@
+import { HmppsAuthClient, OrchestrationApiClient, RestClientBuilder } from '../data'
+
+export default class SupportedPrisonsService {
+  constructor(
+    private readonly orchestrationApiClientFactory: RestClientBuilder<OrchestrationApiClient>,
+    private readonly hmppsAuthClient: HmppsAuthClient,
+  ) {}
+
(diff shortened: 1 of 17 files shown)


server/services/supportedPrisonsService.ts
```diff
@@ -0,0 +1,14 @@
+import { HmppsAuthClient, OrchestrationApiClient, RestClientBuilder } from '../data'
+
+export default class SupportedPrisonsService {
+  constructor(
+    private readonly orchestrationApiClientFactory: RestClientBuilder<OrchestrationApiClient>,
+    private readonly hmppsAuthClient: HmppsAuthClient,
+  ) {}
+
+  async getSupportedPrisonIds(): Promise<string[]> {
+    const token = await this.hmppsAuthClient.getSystemClientToken()
```
tpmcgowan, line 10: In our other apps, we pass `user` to `getSystemClientToken()`, which gets incorporated into the system client token. Works without it though; value of `%ANONYMOUS%` is substituted.

psoleckimoj (approved):
LGTM!

PR #21, opened 2023-09-21
VB-2554 Add integration tests (Cypress / wiremock) for GOV.UK One Login

tpmcgowan:


.circleci/config.yml
```diff
@@ -96,12 +96,18 @@ jobs:
       - restore_cache:
           key: dependency-cache-{{ checksum "package-lock.json" }}
       - run:
-          name: Get wiremock # match version in docker-compose-test.yml
-          command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-jre8-standalone/2.35.0/wiremock-jre8-standalone-2.35.0.jar
+          name: Get wiremock (latest version)
+          command: |
+            LATEST_WIREMOCK_VERSION=$(curl --silent https://repo1.maven.org/maven2/org/wiremock/wiremock-standalone/maven-metadata.xml | sed -n 's/[[:space:]]*<latest>\(.*\)<\/latest>/\1/p')
```
tpmcgowan, line 101: Was stumped for quite a while with a mismatch between Wiremock versions locally and in CircleCI - and breaking changes were making tests fail.

Locally, `docker-compose-test.yml` pulls in the image `wiremock/wiremock` which is (currently) `v3.1.0` whereas CircleCI was hardcoded to use version 2.

What I've added here is scripting to download the latest published standalone Wiremock `.jar` version which should - hopefully - match up with the version published to Docker Hub.

Figured this approach was better than pinning both to a specific version 🤷‍♂️

.circleci/config.yml
```diff
@@ -96,12 +96,18 @@ jobs:
...
+            echo "Wiremock latest version is: $LATEST_WIREMOCK_VERSION"
+            curl -o wiremock.jar "https://repo1.maven.org/maven2/org/wiremock/wiremock-standalone/$LATEST_WIREMOCK_VERSION/wiremock-standalone-$LATEST_WIREMOCK_VERSION.jar"
       - run:
           name: Run wiremock
           command: java -jar wiremock.jar --port 9091
           background: true
+      - run:
+          name: Wait for wiremock to start
```
tpmcgowan, line 109: Because the OIDC Discovery Endpoint needs to be mocked _before_ starting the node app a delay is needed here to make sure Wiremock is available.

integration_tests/mockApis/mappings/openid-configuration.json
```diff
@@ -0,0 +1,48 @@
+{
```
tpmcgowan, line 1: This is a copy of the actual [GOV.UK One Login discovery endpoint data](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/integrate-with-code-flow/#use-the-discovery-endpoint), but with just the hostnames changed.

package.json
```diff
@@ -15,8 +15,9 @@
     "build": "npm run compile-sass && tsc && npm run copy-views",
     "start": "node $NODE_OPTIONS dist/server.js | bunyan -o short",
     "start:dev": "concurrently -k -p \"[{name}]\" -n \"Views,TypeScript,Node,Sass\" -c \"yellow.bold,cyan.bold,green.bold,blue.bold\" \"npm run watch-views\" \"npm run watch-ts\" \"npm run watch-node\" \"npm run watch-sass\"",
-    "start-feature": "export $(cat feature.env) && node $NODE_DEBUG_OPTION dist/server.js | bunyan -o short",
-    "watch-node-feature": "export $(cat feature.env) && nodemon --watch dist/ $NODE_DEBUG_OPTION dist/server.js | bunyan -o short",
+    "oidc-wiremock": "curl --silent --output /dev/null -X POST --data \"@./integration_tests/mockApis/mappings/openid-configuration.json\" http://localhost:9091/__admin/mappings",
+    "start-feature": "npm run oidc-wiremock && export GOVUK_ONE_LOGIN_PRIVATE_KEY=$(cat integration_tests/testKeys/client_private_key.pem) && export $(cat feature.env) && node $NODE_DEBUG_OPTION dist/server.js | bunyan -o short",
```
tpmcgowan, line 19: Before starting the app, need to ensure the OIDC discovery endpoint is mocked. Also need to make sure the private key is in the environment; can't add it to `feature.env` as it didn't work because of the way this is imported.

server/middleware/setUpGovukOneLogin.ts
```diff
@@ -27,11 +27,15 @@ export default function setUpGovukOneLogin(): Router {
     })
 
     router.get('/sign-in', (req, res, next) => {
-      passport.authenticate('oidc', { nonce: generators.nonce(), state: generators.state() })(req, res, next)
+      passport.authenticate('oidc', { nonce: generators.nonce() })(req, res, next)
```
tpmcgowan, line 30: `openid-client` automatically adds `state` parameter if it's left unset.

server/routes/testutils/appSetup.ts
```diff
@@ -64,6 +63,5 @@ export function appWithAllRoutes({
   services?: Partial<Services>
   userSupplier?: () => Express.User
 }): Express {
-  govukOneLogin.default.authenticationMiddleware = () => (req, res, next) => next()
```
tpmcgowan, line 67: Copied this from the way HMPPS Auth was set up but removing as it's actually redundant: we don't `app.use()` the auth so no need to do anything.

server/authentication/govukOneLogin.ts
```diff
@@ -50,7 +50,7 @@ async function init(): Promise<Client> {
   )
 
   const verify: StrategyVerifyCallbackUserInfo<UserinfoResponse> = (tokenSet, userInfo, done) => {
-    logger.info(`GOV.UK One Login user verified: sub =  ${userInfo.sub}`)
+    logger.info(`GOV.UK One Login user verified, sub: ${userInfo.sub}`)
```
psoleckimoj, line 53: Not sure if we want to use `logger.debug()` as I think that only prints it out when on dev 🤔 Probably wrong though.

psoleckimoj (approved):
LGTM!

server/authentication/govukOneLogin.ts
tpmcgowan: I'll have a look at what `logger.debug()` does - haven't experimented with it before. Will sort out what we log (if anything ) about users once we actually start doing something with the returned data.

psoleckimoj (approved):
LGTM!

PR #26, opened 2023-10-30
Update to node 20

psoleckimoj:


package.json
```diff
@@ -107,7 +107,6 @@
     "govuk-frontend": "^4.7.0",
     "helmet": "^7.0.0",
     "http-errors": "^2.0.0",
-    "jwt-decode": "^3.1.2",
```
psoleckimoj, line 110: Removed as can't see it being used.

tpmcgowan: Good spot 🙂! Looks like it's only used in template to get roles from HMPPS Auth token - so shouldn't be needed.

PR #33, opened 2024-02-09
VB-3417 Update from TypeScript template app 2024-02-09

tpmcgowan:


.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ parameters:
 
   node-version:
     type: string
-    default: 20.9-browsers
+    default: 20.11-browsers
```
psoleckimoj, line 30: I wonder if we should stick to `20-browsers` across the board, saves having to do the minor updates all the time 🤔

psoleckimoj (approved):
I'm just going to say LGTM, seems fine but far too many independent changes for me to work out 😂

PR #37, opened 2024-02-23
VB-3475 Update booker home page content

hutcheonb-moj:
## Description

Remove old test api data from home page

Added 'account' url to config  - then to app.locals (part of the govuk one login header)

Added new environment variable to readme


README.md
```diff
@@ -27,6 +27,7 @@ SYSTEM_CLIENT_SECRET="<system_client_secret>"
 ORCHESTRATION_API_URL="https://hmpps-manage-prison-visits-orchestration-dev.prison.service.justice.gov.uk"
 
 GOVUK_ONE_LOGIN_URL=https://oidc.integration.account.gov.uk
+ONE_LOGIN_ACCOUNT_LINK=https://home.integration.account.gov.uk
```
tpmcgowan, line 30: Seeing this group of config items together makes me wonder if the account link should be prefixed with `GOVUK_` for consistency 😀! No need to change though; can tweak some other time if we want...

server/routes/index.ts
```diff
@@ -2,13 +2,15 @@ import { type RequestHandler, Router } from 'express'
...
+import config from '../config'
 
 export default function routes({ supportedPrisonsService }: Services): Router {
   const router = Router()
   const get = (path: string | string[], handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+  const { oneLoginLink } = config
 
   get('/', (req, res, next) => {
-    res.render('pages/index', { user: req.user })
+    res.render('pages/index', { user: req.user, oneLoginLink })
```
tpmcgowan, line 13: I think passing `user` is a hangover from displaying all that debug info so we can remove it at some point...

Passing `oneLoginLink` works here but we'll probably want to eventually move it so it's always available - otherwise we'll have to pass it on every page. For example, could go here (https://github.com/ministryofjustice/hmpps-book-a-prison-visit-ui/blob/c6a863c0152e1c581876d85879ea971a15e945b4/server/utils/nunjucksSetup.ts#L14-L17) then it will always be available in `res.locals`. Fine for this PR though.

tpmcgowan (approved):
LGTM!

tpmcgowan (approved):
LGTM

PR #38, opened 2024-02-27
Use new typography scale

tpmcgowan:
Use new typography scale introduced in frontend v5.2.0: https://github.com/alphagov/govuk-frontend/releases/tag/v5.2.0


hutcheonb-moj (approved):
LGTM

PR #39, opened 2024-02-27
Remove API testing code

tpmcgowan:
Remove code used for initial testing of API calls. Some (mainly testing) code left empty but in place because we will need it.


hutcheonb-moj (approved):
LGTM

PR #42, opened 2024-03-06
VB-3567 Redirect user to originally requested URL after log in

tpmcgowan:
If a user wasn't logged in to GOV.UK One Login and they requested a specific URL within the application they would end up on the home page `"/"` after log in.

This fixes the issue; restoring functionality that was in the HMPPS Auth implementation in the template app.

### DIFF
--- a/integration_tests/index.d.ts
+++ b/integration_tests/index.d.ts
@@ -5,6 +5,6 @@ declare namespace Cypress {
      * Optionally set nonce to override the value used in the ID token
      * @example cy.signIn({ failOnStatusCode: boolean })
      */
-    signIn(options?: { failOnStatusCode: boolean; nonce?: string }): Chainable<unknown>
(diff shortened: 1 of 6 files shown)


integration_tests/index.d.ts
```diff
@@ -5,6 +5,6 @@ declare namespace Cypress {
      * Optionally set nonce to override the value used in the ID token
      * @example cy.signIn({ failOnStatusCode: boolean })
      */
-    signIn(options?: { failOnStatusCode: boolean; nonce?: string }): Chainable<unknown>
+    signIn(options?: { failOnStatusCode: boolean }, nonce?: string, initialRequestUrl?: string): Chainable<unknown>
```
tpmcgowan, line 8: Tweaking this because `options` gets passed to `cy.visit()` (in `integration_tests/support/commands.ts`) and only `failOnStatusCode` is valid for this.

psoleckimoj (approved):
LGTM!

PR #49, opened 2024-05-10
Enable setting GOVUK One Login VTR to low for local / dev environments

tpmcgowan:
Enables setting VTR to a lower level for local, dev and staging environments. This means that users authenticating are not prompted for a second factor code/SMS. (See docs: https://docs.sign-in.service.gov.uk/before-integrating/choose-the-level-of-authentication/#choose-the-level-of-authentication-for-your-service)

Also, some tidying of config for the OL home page URL.


dhirajshettymoj (approved):
LGTM

PR #52, opened 2024-05-14
VB-3725 Select visitors page: add prison-specific visitor config

tpmcgowan:
* Look up prison config to show allowed visitor numbers/ages
* Add GOVUK Error summary component and associated Nunjucks filters for validation errors (copied from Admin UI)
* Update test utils to handle mocking session data and flash
* Add tests for Select visitors page GET and POST route


server/@types/bapv.d.ts
```diff
@@ -9,8 +10,18 @@ export type Booker = {
...
+  prison?: PrisonDto
+
   // all possible visitors for this visit
   allVisitors?: VisitorInfoDto[]
+
   // selected visitors for this visit
   selectedVisitors?: VisitorInfoDto[]
 }
+
+export type FlashData = {
```
tpmcgowan, line 24: For typing what we do with `req.flash()`

server/routes/testutils/appSetup.ts
```diff
@@ -36,15 +36,23 @@ const bookerReference = TestData.bookerReference().value
...
+  production: boolean,
+  userSupplier: () => Express.User,
+  sessionData: SessionData,
+): Express {
   const app = express()
 
   app.set('view engine', 'njk')
 
   nunjucksSetup(app, testAppInfo)
-  app.use(cookieSession({ keys: [''] }))
```
tpmcgowan, line 45: Removing `cookie-session` as it was interfering with simply passing through an object to mock `req.session` - and it didn't appear to be doing anything useful.

server/views/partials/layout.njk
```diff
@@ -26,7 +26,7 @@
   <link href="/assets/stylesheets/application.css?{{ version }}" rel="stylesheet"/>
 {% endblock %}
 
-{% block pageTitle %}{{pageTitle | default(applicationName)}}{% endblock %}
+{% block pageTitle %}{{ pageTitle + " - " if pageTitle }}{{ applicationName }} - GOV.UK{% endblock %}
```
tpmcgowan, line 29: Standardise page `title`s along GDS guidelines

server/routes/testutils/appSetup.ts
psoleckimoj: Where is the session stored if this is removed?

psoleckimoj (approved):
LGTM! Just the question re `cookie-session`

server/routes/testutils/appSetup.ts
tpmcgowan: Basically just a `SessionData` object (defaulting to `{ }`) is now passed through so data can be seeded before requests and stuff can be 'saved' into it. 

I guess without a 'proper' module like `cookie-session` we're not getting any of the things like `session.destroy()` - but nothing like that gets called. (Auth code does, but that's not included in this test app setup; it just assumes a logged in user.)

In Staff UI we removed this too. I think in the Admin UI its still there - but we don't use `req.session` so doesn't matter.

PR #53, opened 2024-05-14
VB-3725 Filter invalid visitor IDs

tpmcgowan:
Add a `customSanitizer()` to the validation chain when selecting visitors to remove any visitor IDs that don't match the list of all visitors.

### DIFF
--- a/server/routes/bookingJourney/selectVisitorsController.ts
+++ b/server/routes/bookingJourney/selectVisitorsController.ts
@@ -52,9 +52,18 @@ export default class SelectVisitorsController {
 
   public validate(): ValidationChain[] {
     return [
-      body('visitorIds').toArray().toInt(),
-      body('visitorIds').isArray({ min: 1 }).withMessage('No visitors selected'),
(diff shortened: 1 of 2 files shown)


server/routes/bookingJourney/selectVisitorsController.ts
```diff
@@ -52,9 +52,18 @@ export default class SelectVisitorsController {
...
   public validate(): ValidationChain[] {
     return [
-      body('visitorIds').toArray().toInt(),
-      body('visitorIds').isArray({ min: 1 }).withMessage('No visitors selected'),
-      // TODO check that it is a valid visitor ID
+      body('visitorIds')
+        .toArray()
+        .toInt()
+        .customSanitizer((visitorIds: number[], meta) => {
+          const req = meta.req as unknown as Express.Request
```
tpmcgowan, line 59: `express-validator` doesn't type `req` fully as an Express request (as it doesn't have to be used with Express). But, we are using express, so it's safe to do this type conversion 😀

PR #62, opened 2024-05-24
VB-3866, hide NOMIS Visitor ID from public UI

hutcheonb-moj:
## Description

Use generated ID for visitor, instead of NOMIS ID

Includes non-working (commented out) method of making test data for visitor


tpmcgowan (approved):
LGTM

PR #66, opened 2024-05-29
VB-3811 Choose date and time - no visit sessions drop-out page

tpmcgowan:
When there are no available visit sessions, show a drop-out page instead of the calendar.


hutcheonb-moj (approved):
LGTM

PR #69, opened 2024-06-03
VB-3941 Update booking journey URLs and refactor code to match

tpmcgowan:


server/routes/bookVisit/additionalSupportController.ts
```diff
@@ -1,16 +1,15 @@
...
 
   public view(): RequestHandler {
     return async (req, res) => {
-      res.render('pages/bookingJourney/selectAdditionalSupport', {
+      res.render('pages/bookVisit/additionalSupport', {
         errors: req.flash('errors'),
         formValues: req.flash('formValues')?.[0] || {},
-        booker: req.session.booker,
-        bookingJourney: req.session.bookingJourney,
+        prisonName: req.session.bookingJourney.prison.prisonName,
```
tpmcgowan, line 12: Unrelated but spotted we were passing unnecessary data to template: only prison name actually used.

server/routes/bookVisit/index.ts
hutcheonb-moj, line 1: Only comment would be should we not change the variable names of the controllers to match their new actual name? selectVisitDateTimeController => chooseVisitTimeController for instance

hutcheonb-moj (dismissed):
LGTM - One suggestion made

server/routes/bookVisit/index.ts
tpmcgowan: Good spot - fixed!

hutcheonb-moj (approved):
LGTM

PR #73, opened 2024-06-11
VB-3944 Update visit application and book visit

tpmcgowan:
* Simplify how main contact data is stored in the booking journey session data (and update test coverage for this)
* call 'change' application endpoint to update the application when the main contact page is submitted
* call 'book' application endpoint when check visit details page submitted and store the relevant visit data ready for the confirmation page

### DIFF
--- a/server/routes/bookVisit/mainContactController.ts
+++ b/server/routes/bookVisit/mainContactController.ts
@@ -1,8 +1,9 @@
 import type { RequestHandler } from 'express'
 import { ValidationChain, body, validationResult } from 'express-validator'
+import { VisitService } from '../../services'
 
 export default class MainContactController {
-  public constructor() {}
+  public constructor(private readonly visitService: VisitService) {}
 
   public view(): RequestHandler {
     return async (req, res) => {
@@ -42,6 +43,8 @@ export default class MainContactController {
       if (req.body.hasPhoneNumber === 'yes') {
         bookingJourney.mainContact.phoneNumber = req.body.phoneNumber
       }
+
+      await this.visitService.changeVisitApplication({ bookingJourney })
 
       return res.redirect('/book-visit/check-visit-details')
     }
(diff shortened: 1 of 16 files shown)


server/routes/bookVisit/mainContactController.ts
```diff
@@ -9,41 +10,39 @@ export default class MainContactController {
...
+        )
+        bookingJourney.mainContact = { contact: contactVisitor }
+      }
 
-      bookingJourney.mainContact = {
-        contact: selectedContact,
-        phoneNumber: req.body.hasPhoneNumber === 'yes' ? req.body.phoneNumber : undefined,
-        contactName: selectedContact === undefined ? req.body.someoneElseName : undefined,
+      if (req.body.hasPhoneNumber === 'yes') {
+        bookingJourney.mainContact.phoneNumber = req.body.phoneNumber
```
hutcheonb-moj, line 45: without having the conditional to set to undefined, when revisiting this page - could you not end up with hasPhoneNumber = no, but booking journey having a phone number, so bad data?

tpmcgowan: Hmmm 🤔 - good question. Just tested out and I think it is OK because on line 34 (or 39), whatever is in `bookingJourney.mainContact` is replaced with a new object so any previous data is gone.

We've got other tickets to look at pre-populating each page in the journey when you go back via 'Check visit' page so these kinds of scenarios will be good to test as part of that.

hutcheonb-moj: good point - replacing the whole object on post should cleanly work around this possibility, so a non-issue in this case

When we re-visit these pages on the pre-populate tickets these are the thing's we have experience testing for anyway so should be good to go

hutcheonb-moj (approved):
LGTM

hutcheonb-moj (approved):
LGTM

PR #75, opened 2024-06-12
VB-3980 Update mod_security rules to stop false positives

tpmcgowan:
* Updates to WAF rules
* rename session cookie

### DIFF
--- a/server/middleware/setUpWebSession.ts
+++ b/server/middleware/setUpWebSession.ts
@@ -20,7 +20,7 @@ export default function setUpWebSession(): Router {
   router.use(
     session({
       store,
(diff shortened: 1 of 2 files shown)


server/middleware/setUpWebSession.ts
```diff
@@ -20,7 +20,7 @@ export default function setUpWebSession(): Router {
   router.use(
     session({
       store,
-      name: 'hmpps-book-a-prison-visit-ui.session',
+      name: 'prison-visits.session',
```
tpmcgowan, line 23: Tweak cookie name in preparation for it featuring on the Cookies page.

psoleckimoj (approved):
LGTM!

PR #76, opened 2024-06-13
VB-3945 Refactor handling of flash in tests

tpmcgowan:
Small refactor to hopefully improve how flash is typed and mocked in tests.

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -28,7 +28,10 @@ export declare global {
 
     interface Request {
       id: string
+      flash(type: 'errors'): ValidationError[]
       flash(type: 'errors', message: ValidationError[]): number
+      flash(type: 'formValues'): Record<string, string | string[]>[]
+      flash(type: 'formValues', message: Record<string, string | string[]>): number
       logout(done: (err: unknown) => void): void
     }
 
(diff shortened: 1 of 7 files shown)


server/routes/testutils/appSetup.ts
```diff
@@ -34,6 +35,12 @@ export const user: Express.User = {
 
 const bookerReference = TestData.bookerReference().value
 
+export type FlashErrors = FieldValidationError[]
```
tpmcgowan, line 38: Moving these type definitions to `appSetup.ts` as they're currently only used in tests.

server/@types/express/index.d.ts
```diff
@@ -28,7 +28,10 @@ export declare global {
 
     interface Request {
       id: string
+      flash(type: 'errors'): ValidationError[]
       flash(type: 'errors', message: ValidationError[]): number
+      flash(type: 'formValues'): Record<string, string | string[]>[]
+      flash(type: 'formValues', message: Record<string, string | string[]>): number
```
tpmcgowan, line 34: Subtlety here (which caused me a lot of confusion!) is that you can set `formValues` with an object,  e.g.:
 ```
req.flash('formValues', req.body) // where req.body = { visitorIds: [1, 2 ] }
```
**But**, when retrieving it you get an array:
```
const formValues = req.flash(`formValues`)
// formValues = [{ visitorIds: [1, 2 ] }]
```

psoleckimoj (approved):
LGTM!

PR #77, opened 2024-06-13
VB-3945 Pre-populate Select visitors page with session data when returning

tpmcgowan:
* When going back to the Select visitors page within a journey, pre-populate it with existing session data
  * New form data overrides this
* Use the `express-validator` `matchedData()` function rather than `req.body` to avoid potentially storing unexpected user data in the session

### DIFF
--- a/server/routes/bookVisit/selectVisitorsController.test.ts
+++ b/server/routes/bookVisit/selectVisitorsController.test.ts
@@ -20,64 +20,64 @@ const prisoner = TestData.prisonerInfoDto()
 const prison = TestData.prisonDto()
 
 const fakeDate = new Date('2024-05-02')
-const visitors = [
-  TestData.visitor({
-    visitorDisplayId: 1,
-    visitorId: 1000,
-    firstName: 'Visitor',
-    lastName: 'Age 20y',
-    dateOfBirth: '2004-04-01',
-  }),
-  TestData.visitor({
-    visitorDisplayId: 2,
-    visitorId: 2000,
-    firstName: 'Visitor',
-    lastName: 'Age 18y',
-    dateOfBirth: '2006-05-02', // 18 today
-  }),
-  TestData.visitor({
-    visitorDisplayId: 3,
-    visitorId: 3000,
-    firstName: 'Visitor',
-    lastName: 'Age 17y',
-    dateOfBirth: '2006-05-03', // 18 tomorrow
-  }),
-  TestData.visitor({
-    visitorDisplayId: 4,
-    visitorId: 4000,
-    firstName: 'Visitor',
-    lastName: 'Age 16y',
-    dateOfBirth: '2008-05-02', // 16 today
-  }),
-  TestData.visitor({
-    visitorDisplayId: 5,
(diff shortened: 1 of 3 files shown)


server/routes/bookVisit/selectVisitorsController.test.ts
```diff
@@ -20,64 +20,64 @@ const prisoner = TestData.prisonerInfoDto()
...
-  TestData.visitor({
-    visitorDisplayId: 8,
-    visitorId: 8000,
-    firstName: 'Visitor',
-    lastName: 'Age 4m',
-    dateOfBirth: '2024-01-02',
-  }),
-]
+
+const visitor1 = TestData.visitor({
```
tpmcgowan, line 24: Just renamed all these so they can be referred to by their ID numbers - rather than mixing that with accessing them from a zero-indexed array. Was getting myself confused 😂

server/routes/bookVisit/selectVisitorsController.ts
```diff
@@ -35,14 +41,15 @@ export default class SelectVisitorsController {
   public submit(): RequestHandler {
     return async (req, res) => {
       const errors = validationResult(req)
+
       if (!errors.isEmpty()) {
         req.flash('errors', errors.array())
-        req.flash('formValues', req.body)
+        req.flash('formValues', matchedData(req, { onlyValidData: false }))
```
tpmcgowan, line 47: Makes sure that only data from expected form fields will end up getting saved with `req.flash()` - i.e. not trusting what's in the request.

(The `onlyValidData: true` property is slightly confusing: it means that it won't return an _unexpected_ field, but would return one that had a legitimate field but with _invalid_ data - e.g. it was too short, for example.)

server/routes/bookVisit/selectVisitorsController.ts
```diff
@@ -35,14 +41,15 @@ export default class SelectVisitorsController {
...
         return res.redirect('/book-visit/select-visitors')
       }
 
       const { bookingJourney } = req.session
-      const { visitorDisplayIds }: { visitorDisplayIds: number[] } = req.body
+      const { visitorDisplayIds } = matchedData<{ visitorDisplayIds: number[] }>(req)
```
tpmcgowan, line 52: Using `matchedData<T>()` to get the valid data - taking advantage of generic return type to define what the validated data will be like.

server/routes/bookVisit/selectVisitorsController.test.ts
psoleckimoj: Is their age all that changes? Wonder if that could be in the var name rather an index? Might be easier to understand what visitors are used for testing rather than referring back to see what age visitor3 is for example?

psoleckimoj (approved):
LGTM!

server/routes/bookVisit/selectVisitorsController.test.ts
tpmcgowan: Good suggestion - I'll have a look at that on a subsequent PR...

PR #78, opened 2024-06-13
VB-3947 Pre-populate Additional support page with session data when returning

tpmcgowan:
* When going back to the Additional support page within a journey, pre-populate it with existing session data - with new form data overriding this


psoleckimoj (approved):
LGTM!

PR #79, opened 2024-06-14
VB-3946 Pre-populate Choose visit time page with session data when returning

tpmcgowan:
* Pre-populate page based on previously selected session date/time
  * if there is an existing application, pass its reference via `excludedApplicationReference` parameter to ensure it is included as an available visit session
* When choosing a time:
  * first time: call 'create application' endpoint to get a new application
  * subsequently: call 'change application' endpoint to update the existing application with the selected time

### DIFF
--- a/server/services/visitService.ts
+++ b/server/services/visitService.ts
@@ -38,11 +38,15 @@ export default class VisitService {
     const token = await this.hmppsAuthClient.getSystemClientToken()
     const orchestrationApiClient = this.orchestrationApiClientFactory(token)
 
-    const { contact, phoneNumber } = bookingJourney.mainContact
-    const visitContact = {
-      name: typeof contact === 'string' ? contact : `${contact.firstName} ${contact.lastName}`,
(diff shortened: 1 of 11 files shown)


server/services/visitService.ts
```diff
@@ -38,11 +38,15 @@ export default class VisitService {
     const token = await this.hmppsAuthClient.getSystemClientToken()
     const orchestrationApiClient = this.orchestrationApiClientFactory(token)
 
-    const { contact, phoneNumber } = bookingJourney.mainContact
-    const visitContact = {
-      name: typeof contact === 'string' ? contact : `${contact.firstName} ${contact.lastName}`,
-      ...(phoneNumber && { telephone: phoneNumber }),
-    }
+    const contact = bookingJourney.mainContact?.contact
```
tpmcgowan, line 41: This method can now be called early in the journey as well as at the end so basically just tweaking it to handle some of the session data not being present.

hutcheonb-moj (approved):
LGTM

PR #81, opened 2024-06-17
VB-3481, visit details page, get route

hutcheonb-moj:
Add allVisitSessions and selectedVisitSession to bookingJourney data object

Display all visit details and back links on 'check visit details' page

### DIFF
--- a/server/routes/bookVisit/checkVisitDetailsController.ts
+++ b/server/routes/bookVisit/checkVisitDetailsController.ts
@@ -8,12 +8,10 @@ export default class CheckVisitDetailsController {
   public view(): RequestHandler {
     return async (req, res) => {
       const { bookingJourney } = req.session
-      let mainContactName = ''
-      if (typeof bookingJourney.mainContact.contact === 'string') {
-        mainContactName = bookingJourney.mainContact.contact
-      } else {
-        mainContactName = `${bookingJourney.mainContact.contact.firstName} ${bookingJourney.mainContact.contact.lastName}`
-      }
+      const mainContactName =
+        typeof bookingJourney.mainContact.contact === 'string'
+          ? bookingJourney.mainContact.contact
+          : `${bookingJourney.mainContact.contact.firstName} ${bookingJourney.mainContact.contact.lastName}`
 
       res.render('pages/bookVisit/checkVisitDetails', {
         additionalSupport: bookingJourney.visitorSupport,
--- a/server/routes/bookVisit/checkVisitDetailsController.ts
+++ b/server/routes/bookVisit/checkVisitDetailsController.ts
@@ -7,9 +7,22 @@ export default class CheckVisitDetailsController {
 
   public view(): RequestHandler {
     return async (req, res) => {
+      const { bookingJourney } = req.session
+      let mainContactName = ''
+      if (typeof bookingJourney.mainContact.contact === 'string') {
+        mainContactName = bookingJourney.mainContact.contact
+      } else {
+        mainContactName = `${bookingJourney.mainContact.contact.firstName} ${bookingJourney.mainContact.contact.lastName}`
+      }
+
       res.render('pages/bookVisit/checkVisitDetails', {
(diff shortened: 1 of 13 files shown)


server/routes/bookVisit/checkVisitDetailsController.ts
```diff
@@ -7,9 +7,22 @@ export default class CheckVisitDetailsController {
 
   public view(): RequestHandler {
     return async (req, res) => {
+      const { bookingJourney } = req.session
+      let mainContactName = ''
+      if (typeof bookingJourney.mainContact.contact === 'string') {
+        mainContactName = bookingJourney.mainContact.contact
+      } else {
+        mainContactName = `${bookingJourney.mainContact.contact.firstName} ${bookingJourney.mainContact.contact.lastName}`
+      }
```
tpmcgowan, line 16:
```suggestion
const mainContactName =
        typeof bookingJourney.mainContact.contact === 'string'
          ? bookingJourney.mainContact.contact
          : `${bookingJourney.mainContact.contact.firstName} ${bookingJourney.mainContact.contact.lastName}`
```

server/routes/bookVisit/chooseVisitTimeController.ts
```diff
@@ -61,8 +61,14 @@ export default class ChooseVisitTimeController {
...
 
       const { booker, bookingJourney } = req.session
-      bookingJourney.selectedSessionDate = selectedSessionDate
-      bookingJourney.selectedSessionTemplateReference = selectedSessionTemplateReference
+
+      const selectedVisitSession = bookingJourney.allVisitSessions.find(
+        singleVisitSession =>
+          singleVisitSession.sessionTemplateReference.includes(selectedSessionTemplateReference) &&
+          singleVisitSession.sessionDate.includes(selectedSessionDate),
+      )
```
tpmcgowan, line 69:
```suggestion

      const selectedVisitSession = bookingJourney.allVisitSessions.find(
        session =>
          session.sessionTemplateReference === selectedSessionTemplateReference &&
          session.sessionDate === selectedSessionDate,
      )
```

We're just comparing strings here so no need for `.includes()` (which is for checking if a value is in an array)

server/views/pages/bookVisit/checkVisitDetails.njk
```diff
@@ -1,22 +1,153 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h1 class="govuk-heading-l">Check the visit details before booking</h1>
+
+      {% include "partials/errorSummary.njk" %}
+
+      {% set additionalSupportRows %}
+      {% if additionalSupport | length %}{{ additionalSupport }}{% else %}None{% endif %}
+      {% endset %}
```
tpmcgowan, line 18: This looks like a hangover from when additional support was more than just a text field. Suggest removing this and changing line 97 to:
```
html: '<p class="test-additional-support">' + (additionalSupport if additionalSupport | length else "None") + '</p>'
```

tpmcgowan (commented):
Looks good 👍 . A few minor comments

server/utils/nunjucksSetup.ts
```diff
@@ -89,5 +89,9 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
 
   njkEnv.addFilter('pluralise', pluralise)
 
+  njkEnv.addFilter('formatTelephone', (telephoneNumber: string) => {
+    return telephoneNumber === 'UNKNOWN' || telephoneNumber === undefined ? 'No phone number provided' : telephoneNumber
+  })
+
```
tpmcgowan, line 95: I don't think we need this as we shouldn't have to handle numbers with `UNKNOWN` - we'll only be showing visits that originated in this service.

So, suggest just putting in the template file:
```
...  mainContactNumber | d("No phone number provided")   ...
```

tpmcgowan (approved):
LGTM!

PR #84, opened 2024-06-18
VB-3861 Add session validation middleware for booking journey

tpmcgowan:
Two main changes:
* Add a new `paths` constant to hold route paths. This avoids them being repeated across code and tests. It's also added to `app.locals` so is available in templates.
* New `bookVisitSessionValidator` middleware to check the necessary is in the user's session at each stage in the journey

Also test updates to cover the above.

### DIFF
--- a/server/constants/paths.ts
+++ b/server/constants/paths.ts
@@ -1,0 +1,16 @@
+const paths = {
+  HOME: '/',
+
+  BOOK_VISIT: {
+    ROOT: '/book-visit',
+    SELECT_PRISONER: '/book-visit/select-prisoner',
+    SELECT_VISITORS: '/book-visit/select-visitors',
+    CHOOSE_TIME: '/book-visit/choose-visit-time',
+    ADDITIONAL_SUPPORT: '/book-visit/additional-support',
+    MAIN_CONTACT: '/book-visit/main-contact',
+    CHECK_DETAILS: '/book-visit/check-visit-details',
+    BOOKED: '/book-visit/visit-booked',
+  },
+} as const
+
+export default paths
--- a/server/middleware/bookVisitSessionValidator.ts
+++ b/server/middleware/bookVisitSessionValidator.ts
@@ -1,0 +1,60 @@
+import { RequestHandler, Response } from 'express'
+import createError from 'http-errors'
+import paths from '../constants/paths'
+import logger from '../../logger'
+
+const journeyOrder: string[] = [
+  paths.BOOK_VISIT.SELECT_PRISONER,
+  paths.BOOK_VISIT.SELECT_VISITORS,
+  paths.BOOK_VISIT.CHOOSE_TIME,
+  paths.BOOK_VISIT.ADDITIONAL_SUPPORT,
+  paths.BOOK_VISIT.MAIN_CONTACT,
+  paths.BOOK_VISIT.CHECK_DETAILS,
+  paths.BOOK_VISIT.BOOKED,
+]
+
+export default function bookVisitSessionValidator(): RequestHandler {
+  return (req, res, next) => {
+    const { booker, bookingJourney } = req.session
+    const { baseUrl, method, path } = req
+    const requestPath = baseUrl + path
+
+    const journeyStage = journeyOrder.indexOf(requestPath)
+
+    if (journeyStage === -1) {
+      return next(createError(404, 'Not found'))
+    }
+
+    if (!booker.prisoners?.length) {
+      return logAndRedirect(res, method, requestPath, booker.reference)
+    }
+
+    if (journeyStage >= journeyOrder.indexOf(paths.BOOK_VISIT.SELECT_VISITORS) && !bookingJourney.prisoner) {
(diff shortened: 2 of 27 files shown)


server/middleware/bookVisitSessionValidator.ts
```diff
@@ -0,0 +1,67 @@
+import { RequestHandler, Response } from 'express'
+import createError from 'http-errors'
+import paths from '../constants/paths'
+import logger from '../../logger'
+
+const journeyOrder: string[] = [
```
tpmcgowan, line 6: Idea here is to set journey order in an array so the numerical 'stage' in the journey can be calculated by comparing the array indices of the current & test paths.

So, quite similar to what `stage` is doing in the equivalent Staff UI middleware ([server/middleware/sessionCheckMiddleware.ts](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/main/server/middleware/sessionCheckMiddleware.ts)) but here the middleware calculates the stage itself rather than having it passed in per route. I'm hoping this might be simpler to maintain if routes are added/changed.

server/constants/paths.ts
```diff
@@ -0,0 +1,16 @@
+const paths = {
```
tpmcgowan, line 1: Most of the rest of the changes in this PR are just updating everywhere else to use these constant values instead of hard-coded strings.

server/routes/bookVisit/additionalSupportController.test.ts
```diff
@@ -46,6 +50,17 @@ describe('Additional support needs', () => {
 
       app = appWithAllRoutes({ sessionData })
     })
+    it('should use the session validation middleware', () => {
```
tpmcgowan, line 53: Added a similar test to each `/book-visit` route to be sure the validation middleware actually being called.

server/routes/index.ts
```diff
@@ -12,9 +13,9 @@ export default function routes(services: Services): Router {
 
   const home = new HomeController(services.bookerService)
 
-  get('/', home.view())
+  get(paths.HOME, home.view())
 
-  router.use('/book-visit', bookingJourneyRoutes(services))
+  router.use(bookingJourneyRoutes(services))
```
tpmcgowan, line 18: Not setting path to `/book-visit` when mounting this route handler because the full path is specified when mounting the individual page routes in `server/routes/bookVisit/index.ts`. Keeps the `paths` constant simpler (i.e. I couldn't figure out an alternative set up that didn't require storing both the full paths **and** all the segments 😅!)

server/routes/bookVisit/index.ts
```diff
@@ -27,29 +29,37 @@ export default function routes(services: Services): Router {
   const checkVisitDetailsController = new CheckVisitDetailsController(services.visitService)
   const visitBookedController = new VisitBookedController()
 
-  // TODO need session checks for each stage to validate what is in session - add middleware here to apply to all booking journey routes?
+  router.use(paths.BOOK_VISIT.ROOT, bookVisitSessionValidator())
```
tpmcgowan, line 32: Applies the validation middleware to any path matching `/book-visit`.

server/middleware/bookVisitSessionValidator.ts
```diff
@@ -0,0 +1,100 @@
...
+  paths.BOOK_VISIT.MAIN_CONTACT,
+  paths.BOOK_VISIT.CHECK_DETAILS,
+  paths.BOOK_VISIT.BOOKED,
+]
+
+export default function bookVisitSessionValidator(): RequestHandler {
+  return (req, res, next) => {
+    const { booker, bookingJourney, bookingConfirmed } = req.session
+    const { baseUrl, method, path } = req
+    const requestPath = baseUrl + path
```
psoleckimoj, line 20: `${baseUrl}${path}` 😛

server/middleware/bookVisitSessionValidator.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const journeyStage = journeyOrder.indexOf(requestPath)
+
+    // Unknown path
+    if (journeyStage === -1) {
+      return next(createError(404))
+    }
+
+    // No booker prisoner - any path
+    if (!booker.prisoners?.length) {
```
psoleckimoj, line 30: Could you maybe combine all the conditions? Just seems 90% of them actually return the same thing?

psoleckimoj (approved):
LGTM!

server/middleware/bookVisitSessionValidator.ts
tpmcgowan: Surprised that TypeScript / linter didn't complain about that. Fixed in upcoming PR...

server/middleware/bookVisitSessionValidator.ts
tpmcgowan: Hmm, yes, worth a look 🤔. We've got a couple of tickets next to add some more paths (for various drop-out points) so when I wedge those in I can see if it can be consolidated a bit.

PR #85, opened 2024-06-19
VB-3953, VB-3777, VB-3951, VB-3952 Handle failed application create/change and booking

tpmcgowan:
If there is an `400 Bad Request` API error during application create / change / book this means that the requested visit session is no longer available. Handle this by returning to the calendar and showing a message.

Any other API errors are propagated to be handled as the application does by default.

Screenshot 2024-06-19 at 12 40 08


assets/scss/components/moj/banner/_banner.scss
```diff
@@ -0,0 +1,55 @@
+/* ==========================================================================
```
tpmcgowan, line 1: The design uses the Banner component from the MoJ Design System. As it is such a simple component - and I think the only one we're planning - I have just copied it.

I figured this would be preferable to adding the whole MoJ library in - which we could do if we end up using more components... (Or we could just add it now - happy to change 🤷‍♂️ )

server/routes/bookVisit/checkVisitDetailsController.ts
```diff
@@ -43,7 +43,12 @@ export default class CheckVisitDetailsController {
         delete req.session.bookingJourney
         return res.redirect(paths.BOOK_VISIT.BOOKED)
       } catch (error) {
-        // TODO handle errors (VB-3597)
+        // HTTP 400 Bad Request is the response when a session is no longer available
+        if (error.status === 400) {
+          req.flash('message', 'Your visit time is no longer available. Select a new time.')
```
tpmcgowan, line 48: This string is duplicated in a few places but this will be handled better when text moved into the il8n framework (coming soon...)

assets/scss/components/moj/banner/_banner.scss
psoleckimoj: I can see from both sides.

Including only this component means no need to import the library and set that up, but I'd make sure everywhere that code is used from that library needs a comment so it's easy to find when/if it gets replaced? The negative is (as unlikely as it is), if the component ever gets updated we won't get those updates so I'm happy to leave to you to decide 😁

psoleckimoj (approved):
LGTM!

assets/scss/components/moj/banner/_banner.scss
tpmcgowan: Updated to include the MoJ library just for the component we use (so far). Haven't bothered adding the JS initialisation bits as the banner component doesn't use this so no benefit at this stage.

psoleckimoj (approved):
LGTM!

PR #88, opened 2024-06-21
VB-3805, VB-3833 Add 'Visit cannot be booked' drop-out page and check VO count at booking journey start 

tpmcgowan:
* Incorporate API changes to Booker / Prisoner data format
* New 'A visit cannot be booked' drop out page for VO's that gives prisoner name and next VO issue date
* Catch prisoner having no VOs at start of booking journey and redirect to this page.

### DIFF
--- a/server/data/orchestrationApiTypes.ts
+++ b/server/data/orchestrationApiTypes.ts
@@ -16,7 +16,7 @@ export type CreateApplicationDto = components['schemas']['CreateApplicationDto']
 
 export type PrisonDto = components['schemas']['PrisonDto']
 
-export type PrisonerInfoDto = components['schemas']['PrisonerInfoDto']
+export type BookerPrisonerInfoDto = components['schemas']['BookerPrisonerInfoDto']
 
 export type VisitDto = components['schemas']['VisitDto']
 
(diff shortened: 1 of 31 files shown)


server/data/orchestrationApiTypes.ts
```diff
@@ -16,7 +16,7 @@ export type CreateApplicationDto = components['schemas']['CreateApplicationDto']
 
 export type PrisonDto = components['schemas']['PrisonDto']
 
-export type PrisonerInfoDto = components['schemas']['PrisonerInfoDto']
+export type BookerPrisonerInfoDto = components['schemas']['BookerPrisonerInfoDto']
```
tpmcgowan, line 19: Most of the changes in this PR are a result of this DTO changing name and structure. I've kept the changes to the data layer; in the service layer now fully define our own `Prisoner` type with the data we need and a better structure.

server/services/bookerService.ts
```diff
@@ -1,10 +1,16 @@
 import logger from '../../logger'
 import { HmppsAuthClient, OrchestrationApiClient, RestClientBuilder } from '../data'
-import { AuthDetailDto, PrisonerInfoDto, VisitorInfoDto } from '../data/orchestrationApiTypes'
+import { AuthDetailDto, VisitorInfoDto } from '../data/orchestrationApiTypes'
 import { isAdult } from '../utils/utils'
 
-export interface Prisoner extends PrisonerInfoDto {
+export type Prisoner = {
```
tpmcgowan, line 6: Defining `Prisoner` here rather than extending the type from the back-end. This is to give a flat structure and only the properties that are used.

psoleckimoj (approved):
LGTM!

hutcheonb-moj (approved):
LGTM

PR #89, opened 2024-06-24
VB-3851, VB-3885 Set up static content pages for viewing when not logged in

tpmcgowan:
* Set up a set of routes that are loaded before `authenticate()` middleware run so these pages can be viewed by both authenticated and unauthenticated users
* Configure page layout to show GOVUK One Login header for logged in users and fallback header for others
* Unit and integration tests

### DIFF
--- a/server/middleware/setUpGovukOneLogin.ts
+++ b/server/middleware/setUpGovukOneLogin.ts
@@ -47,8 +47,6 @@ export default function setUpGovukOneLogin(): Router {
       } else res.redirect(client.endSessionUrl())
     })
 
-    router.use(govukOneLogin.authenticationMiddleware())
-
     router.use((req, res, next) => {
       res.locals.user = req.user
       next()
--- a/server/routes/unauthenticatedRoutes.ts
+++ b/server/routes/unauthenticatedRoutes.ts
@@ -1,0 +1,14 @@
+import { Router } from 'express'
+
+import paths from '../constants/paths'
+
+export default function routes(): Router {
+  const router = Router()
+
+  // Accessibility statement
+  router.get(paths.ACCESSIBILITY, (req, res) => {
+    res.render('pages/accessibilityStatement')
+  })
+
+  return router
+}
--- a/server/views/pages/accessibilityStatement.njk
+++ b/server/views/pages/accessibilityStatement.njk
@@ -1,0 +1,13 @@
+{% extends "partials/layout.njk" %}
+
+{% set pageTitle = "Accessibility statement" %}
+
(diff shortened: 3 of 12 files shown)


server/middleware/setUpGovukOneLogin.ts
```diff
@@ -47,8 +47,6 @@ export default function setUpGovukOneLogin(): Router {
       } else res.redirect(client.endSessionUrl())
     })
 
-    router.use(govukOneLogin.authenticationMiddleware())
```
tpmcgowan, line 50: Moving this to be called in `app.ts` so routes can be called before/after it to allow for unauthenticated access

server/routes/unauthenticatedRoutes.ts
```diff
@@ -0,0 +1,14 @@
+import { Router } from 'express'
+
+import paths from '../constants/paths'
+
+export default function routes(): Router {
```
tpmcgowan, line 5: Haven't bothered with individual controllers for the various 'accessibility', 'privacy' etc pages as they're so simple.

server/views/pages/accessibilityStatement.njk
```diff
@@ -0,0 +1,13 @@
+{% extends "partials/layout.njk" %}
+
+{% set pageTitle = "Accessibility statement" %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h1 class="govuk-heading-l">Accessibility statement</h1>
```
tpmcgowan, line 9: I'll put the actual content of this (and the other static pages) in next PR - this one is just to get the mechanism of them working in.

server/routes/unauthenticatedRoutes.ts
```diff
@@ -0,0 +1,14 @@
...
+  const router = Router()
+
+  // Accessibility statement
+  router.get(paths.ACCESSIBILITY, (req, res) => {
+    res.render('pages/accessibilityStatement')
```
psoleckimoj, line 10: This is where you wonder if the path should map to the rendered page template as well, and maybe a type then you could just loop round them passing in path and page. But that's a possible for another day 😁

psoleckimoj (approved):
LGTM!

server/routes/unauthenticatedRoutes.ts
tpmcgowan: Yeah, that could be good - will resist the temptation to refactor it all now though unless/until we get a lot of these static pages!

PR #96, opened 2024-06-25
VB-3783, VB-3944 Disable submit buttons on submit (create & book)

tpmcgowan:
Disable submit buttons when clicked at key points in the journey to avoid the risk of 'double bookings', e.g. if user attempts to re-try because response is slow.

Applies to: choosing slot on calendar page (i.e. visit application creation) and check details page (i.e. visit booking).

Based on similar changes to Staff UI (https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/pull/766)


psoleckimoj (approved):
LGTM!

PR #100, opened 2024-06-27
VB-3479 Refactor and tighten main contact validation

tpmcgowan:
Refactor and improve validation on Main contact page:
* Expand test data
* Check that when selecting a visitor as a main contact that it is both a valid selected visitor ID and  an adult visitor
* Refactor `express-validator` code to be simpler and use more built-in functions

### DIFF
--- a/server/routes/bookVisit/mainContactController.test.ts
+++ b/server/routes/bookVisit/mainContactController.test.ts
@@ -19,8 +19,14 @@ let sessionData: SessionData
 const bookerReference = TestData.bookerReference().value
 const prisoner = TestData.prisoner()
 const prison = TestData.prisonDto()
-const adultVisitor = TestData.visitor()
-const childVisitor = TestData.visitor({ dateOfBirth: `${new Date().getFullYear() - 2}-01-01`, adult: false })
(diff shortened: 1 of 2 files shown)


server/routes/bookVisit/mainContactController.test.ts
```diff
@@ -19,8 +19,14 @@ let sessionData: SessionData
 const bookerReference = TestData.bookerReference().value
 const prisoner = TestData.prisoner()
 const prison = TestData.prisonDto()
-const adultVisitor = TestData.visitor()
-const childVisitor = TestData.visitor({ dateOfBirth: `${new Date().getFullYear() - 2}-01-01`, adult: false })
+const adultVisitor1 = TestData.visitor({ visitorDisplayId: 1, visitorId: 100 })
```
tpmcgowan, line 22: Making sure the test contacts have different IDs (otherwise they are all defaulting to `1`). Add another contact so it can be confirmed that main contact choice is only from `selectedVisitors`, which is a subset of `allVisitors`.

psoleckimoj (approved):
LGTM!

PR #110, opened 2024-07-05
VB-3918 future visit single page

hutcheonb-moj:
Visit details page for future visits


server/constants/prisonInformation.ts
```diff
@@ -0,0 +1,30 @@
...
+    prisonName: 'Drake Hall (HMP)',
+    prisonPhoneNumber: '0121 661 2101',
+    prisonWebsite: 'https://www.gov.uk/guidance/drake-hall-prison',
+  },
+  FHI: {
+    prisonName: 'Foston Hall (HMP)',
+    prisonPhoneNumber: '0121 661 2101',
+    prisonWebsite: 'https://www.gov.uk/guidance/foston-hall-prison',
+  },
+} as const
```
psoleckimoj, line 18: Do you need this when it's declared as a `const`?

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -0,0 +1,48 @@
...
+
+      // if manual number entered (larger than current number of visits in session data)
+      // redirect to bookings page
+      if (visitNumber > bookings.length) {
+        return res.redirect(paths.BOOKINGS.HOME)
+      }
+
+      const prisoner = booker.prisoners[0]
+
+      const { prisonName, prisonPhoneNumber, prisonWebsite } = getPrisonInformation(prisoner.prisonCode)
```
psoleckimoj, line 31: Is this not available from any APIs anywhere? Just wondering if we could get this from a central place as we don't really want an app to be the source of truth for these details?

hutcheonb-moj: it was agreed that for now we'd hard code these - thats why i've added them into a constant file (we were duplicating it across places)

server/constants/prisonInformation.ts
hutcheonb-moj: just following practices used in Staff UI (has same method, same way of definition etc)

server/routes/testutils/testData.ts
```diff
@@ -126,7 +126,7 @@ export default class TestData {
     endTimestamp = '2024-05-30T11:30:00',
     visitNotes = [],
     visitContact = { name: 'Joan Phillips', telephone: '01234 567890' },
-    visitors = [{ nomisPersonId: 1234, visitContact: true }],
+    visitors = [{ nomisPersonId: 1234, visitContact: true, firstName: 'Keith', lastName: 'Phillips' }],
```
tpmcgowan, line 129: Is this test data logically correct? if `visitors[0]` has `visitContact: true` then the name should match what's in `visitContact`?

server/constants/prisonInformation.ts
tpmcgowan: Probably me that originally put the `as const` in 😅 ! If I remember correctly, I *think* it stops the data being modified. E.g. if you have `const name = 'Tom'` then you can't change it. But if you had `const person = { name: 'Tom' }`, you can't change the whole object but you could do `person.name = 'xyz'` - which we wouldn't want.

server/views/pages/bookings/visit.njk
```diff
@@ -0,0 +1,92 @@
+{% set activeLink = 'bookings-link' %}
+{% extends "../../partials/layout.njk" %}
+
+{% set pageTitle = "Bookings" %}
```
tpmcgowan, line 4:
```suggestion
{% set pageTitle = "Visit booking details" %}
```

... perhaps then use this variable in line 13 for the `h1` (in fact we should probably do this elsewhere as a searate tweak as it looks like we're duplicating the `title` and `h1` text in most of the other templates!)

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -0,0 +1,48 @@
...
+        prisonName,
+        prisonPhoneNumber,
+        prisonWebsite,
+        showServiceNav: true,
+      })
+    }
+  }
+
+  public validate(): ValidationChain[] {
+    return [param('visitNumber').exists().notEmpty().toInt().isInt({ min: 1 })]
```
tpmcgowan, line 46:
```suggestion
    return [param('visitNumber').toInt().isInt({ min: 1 })]
```
I think this should achieve the same result with fewer tests.

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -0,0 +1,48 @@
...
+export default class BookingDetailsController {
+  public constructor(private readonly visitService: VisitService) {}
+
+  public view(): RequestHandler {
+    return async (req, res) => {
+      const { booker, bookings } = req.session
+
+      const errors = validationResult(req)
+      if (!errors.isEmpty()) {
+        req.flash('errors', errors.array())
```
tpmcgowan, line 16:
```suggestion
```
Don't want to set flash here as we don't load it where you're redirecting to. This means it could randomly appear on the next page they hit where `flash('errors')` is read!

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -0,0 +1,48 @@
...
+        return res.redirect(paths.BOOKINGS.HOME)
+      }
+
+      const visitNumber = +req.params.visitNumber
```
tpmcgowan, line 20:
```suggestion
      const { visitNumber } = matchedData<{ visitNumber: number }>(req)
```

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -0,0 +1,48 @@
...
+      const visit = bookings[visitNumber - 1]
+
+      // if manual number entered (larger than current number of visits in session data)
+      // redirect to bookings page
+      if (visitNumber > bookings.length) {
+        return res.redirect(paths.BOOKINGS.HOME)
+      }
```
tpmcgowan, line 27: Might want to re-order to check first if visitNumber is going to be out of bounds for the array. E.g. currently it will try to access `bookings[999]` and assign to `visit` **before** checking if this is a valid value.

server/routes/bookings/bookingDetailsController.test.ts
```diff
@@ -0,0 +1,70 @@
...
+})
+
+describe('View single booking', () => {
+  it('should render the booking details page', () => {
+    return request(app)
+      .get(`${paths.BOOKINGS.VISIT}/1`)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        const $ = cheerio.load(res.text)
+        expect($('title').text()).toMatch(/^Bookings -/)
```
tpmcgowan, line 42:
```suggestion
        expect($('title').text()).toMatch(/^Visit booking details -/)
```

server/routes/bookings/bookingDetailsController.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        expect($('[data-test="back-link"]').length).toBe(1)
```
tpmcgowan, line 43:
```suggestion
        expect($('[data-test="back-link"]').attr('href')).toBe(paths.BOOKINGS.HOME)
```
Should check that the link is correct; not just that it exists.

integration_tests/e2e/futureVisits.cy.ts
```diff
@@ -30,5 +35,18 @@ context('Bookings home page', () => {
     bookingsPage.visitStartTime(1).contains('10am')
     bookingsPage.visitEndTime(1).contains('11:30am')
     bookingsPage.visitReference(1).contains('ab-cd-ef-gh')
+
+    bookingsPage.visitLink(1).click()
```
tpmcgowan, line 39: GH won't let me comment on line 22, but now that the test has been extended, perhaps update the description. E.g.:
```
it('should show bookings home page with a future visit and navigate to view the visit details', () => {
```
... or similar 🤷‍♂️

tpmcgowan (approved):
LGTM!

PR #120, opened 2024-07-12
VB-4183 Update CSP 'formAction' to include GOV.UK One Login URL

tpmcgowan:
Fixes an issue where if a user had an expired session then tried a form submission it would fail - and appear to 'hang' - because the resultant redirect to re-authenticate with GOV.UK One Login would be blocked by the browser for violating the CSP policy.

### DIFF
--- a/server/authentication/govukOneLogin.ts
+++ b/server/authentication/govukOneLogin.ts
@@ -19,6 +19,7 @@ passport.deserializeUser((user, done) => {
 const authenticationMiddleware = (): RequestHandler => {
   return async (req, res, next) => {
     if (req.isAuthenticated()) {
+      res.locals.user = req.user
       return next()
     }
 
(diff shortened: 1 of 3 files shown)


server/authentication/govukOneLogin.ts
```diff
@@ -19,6 +19,7 @@ passport.deserializeUser((user, done) => {
 const authenticationMiddleware = (): RequestHandler => {
   return async (req, res, next) => {
     if (req.isAuthenticated()) {
+      res.locals.user = req.user
```
tpmcgowan, line 22: Unrelated change - just moving code to populate `res.locals.user` to make sure it happens if the user has actually been authenticated. Basically a tidy up after moving authentication middleware in #89 to support some routes not requiring authentication.

hutcheonb-moj (approved):
LGTM

PR #123, opened 2024-07-15
VB-4183 CSP form-action fix

tpmcgowan:
Add '`signin`' as well as the '`oidc`' sub-domain for the GOV.UK One Login URL to the CSP `form-action` rule.

Needed because with an expired session, when a user makes a `POST` request, Chrome blocks the initial redirect to the `oidc` domain and then the subsequent redirect to the `signin` domain.


hutcheonb-moj (approved):
LGTM

PR #125, opened 2024-07-16
VB-4185 booking details refresh bug

hutcheonb-moj:
Both pages are now available after session refresh / timeout


server/routes/bookings/bookingsController.ts
```diff
@@ -1,19 +1,24 @@
...
+  public constructor(
+    private readonly visitService: VisitService,
+    private readonly bookerService: BookerService,
+  ) {}
 
   public view(): RequestHandler {
     return async (req, res) => {
       const { booker } = req.session
 
+      const prisoner = booker.prisoners ? booker.prisoners : await this.bookerService.getPrisoners(booker.reference)
```
tpmcgowan, line 15: Might be good to cover this new dependency in `bookingsController.test.ts`. E.g.:
* in the current test, assert that `getPrisoners()` is **not** called
* duplicate the test but with `booker.prisoners` not set and assert that `getPrisoners()` is called as expected (this test could be minimal; i.e. don't need to re-check all the page elements)

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -1,21 +1,32 @@
...
         return res.redirect(paths.BOOKINGS.HOME)
       }
 
+      req.session.bookings = req.session.bookings
+        ? req.session.bookings
+        : await this.visitService.getFuturePublicVisits(booker.reference)
+
+      const prisoner = booker.prisoners ? booker.prisoners : await this.bookerService.getPrisoners(booker.reference)
+
+      const { bookings } = req.session
```
tpmcgowan, line 28:
```suggestion
      const { booker, bookings } = req.session

      const errors = validationResult(req)
      if (!errors.isEmpty() || !bookings?.length) {
        return res.redirect(paths.BOOKINGS.HOME)
      }
```

Something like this suggestion (caveat - untested!) and then keeping the rest as it was might be preferable because:
* simpler - avoids duplicating the code to fetch future bookings
* less potential for confusion because if your session expires on a visit details page you just get returned to the listing. If you went back to, say, `.../1` then in the meantime the another visit could have been booked so your page would reload with info on a **different** visit

server/routes/bookings/bookingsController.test.ts
```diff
@@ -75,4 +75,19 @@ describe('Bookings homepage', () => {
         expect($('h2').text()).not.toContain('How to change your booking')
       })
   })
+
+  it('should render the bookings home page - if no prisoner was present on initial call', () => {
+    sessionData.booker.prisoners = undefined
+    visitService.getFuturePublicVisits.mockResolvedValue([orchestrationVisitDto])
+    bookerService.getPrisoners.mockResolvedValue([prisoner])
```
tpmcgowan, line 82: Probably no longer need this test now that the behaviour of the route has changed (or at least `bookerService` mock can go as it's no longer relevant)

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -1,21 +1,23 @@
...
     return async (req, res) => {
       const { booker, bookings } = req.session
 
       const errors = validationResult(req)
-      if (!errors.isEmpty()) {
+      if (!errors.isEmpty() || !bookings?.length) {
         return res.redirect(paths.BOOKINGS.HOME)
       }
 
+      const prisoner = booker.prisoners ? booker.prisoners : await this.bookerService.getPrisoners(booker.reference)
```
tpmcgowan, line 19:
```suggestion
      const prisoner = booker.prisoners?.[0] ?? (await this.bookerService.getPrisoners(booker.reference))?.[0]
```

(and on line 34, just `prisoner,` instead of `prisoner: prisoner[0],`)

This is because it's still possible for `prisoner` to be empty. Also, keeps naming more appropriate as `prisoner` will be a `Prisoner | undefined` rather than `Prisoner[]`

Ultimately, if there is no prisoner it will go to template as `undefined` but tested and Nunjucks handles this without error (though I guess up to Simon if we display a message rather than just nothing 🤷‍♂️)

tpmcgowan (approved):
LGTM

PR #127, opened 2024-07-17
VB-4196 Remove past/cancelled links

tpmcgowan:
Temporarily remove the links to view past/cancelled visits until these pages are built.


hutcheonb-moj (approved):
LGTM

PR #128, opened 2024-07-17
VB-3580 Handle book application error response codes (no VOs or no capacity)

tpmcgowan:
* Update to handle potential HTTP `422` response and validation error codes when making the API call to 'book' an application 
* Small refactor to split up some integration tests to separate files

Also:
* fix issue with when `res.locals.user` is populated
* update content on error page
* minor tweaks to paths and link visited states

### DIFF
--- a/integration_tests/e2e/bookingJourney.cy.ts
+++ b/integration_tests/e2e/bookingJourney.cy.ts
@@ -4,14 +4,12 @@ import { DateFormats } from '../../server/constants/dateFormats'
 import TestData from '../../server/routes/testutils/testData'
 import AdditionalSupportPage from '../pages/bookVisit/additionalSupport'
 import ChooseVisitTimePage from '../pages/bookVisit/chooseVisitTime'
-import ChooseVisitTimeNoSessionsPage from '../pages/bookVisit/chooseVisitTimeNoSessions'
 import HomePage from '../pages/home'
 import Page from '../pages/page'
 import SelectVisitorsPage from '../pages/bookVisit/selectVisitors'
 import MainContactPage from '../pages/bookVisit/mainContact'
 import CheckVisitDetailsPage from '../pages/bookVisit/checkVisitDetails'
 import VisitBookedPage from '../pages/bookVisit/visitBooked'
-import CannotBookPage from '../pages/bookVisit/cannotBook'
 
 context('Booking journey', () => {
   const today = new Date()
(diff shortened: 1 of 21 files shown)


integration_tests/e2e/bookingJourney.cy.ts
tpmcgowan, line 1: No actual change to the Cypress tests - just splitting out the booking journey tests and the 'drop-out' points tests to separate files to simplify.

package.json
```diff
@@ -162,6 +162,7 @@
     "mocha-junit-reporter": "^2.2.1",
     "nock": "^13.5.4",
     "nodemon": "^3.1.4",
+    "openapi-typescript": "^7.0.3",
```
tpmcgowan, line 165: Hopefully pinning version here will stop us having conflicting versions of the API definitions!

server/routes/bookVisit/checkVisitDetailsController.ts
```diff
@@ -47,12 +48,20 @@ export default class CheckVisitDetailsController {
         delete req.session.bookingJourney
         return res.redirect(paths.BOOK_VISIT.BOOKED)
       } catch (error) {
-        // HTTP 400 Bad Request is the response when a session is no longer available
-        if (error.status === 400) {
-          req.flash('message', 'Your visit time is no longer available. Select a new time.')
-          delete bookingJourney.selectedVisitSession
-          return res.redirect(paths.BOOK_VISIT.CHOOSE_TIME)
+        if (error.status === 422 && error.validationErrors?.length) {
```
tpmcgowan, line 51: It's likely that these checks and new error code responses will be added to the create/change application in the future. When they are, these conditions can be moved to a separate helper function to avoid duplication.

server/sanitisedError.ts
```diff
@@ -1,18 +1,18 @@
 import type { ResponseError } from 'superagent'
 
-export interface SanitisedError extends Error {
+export interface SanitisedError<Data = unknown> extends Error {
```
tpmcgowan, line 3: Using TS generics here so we can pass through a type for `data` (and with `unknown` as a default). Means we get TS support for the error code values in the error response.

server/routes/bookVisit/checkVisitDetailsController.ts
```diff
@@ -47,12 +49,26 @@ export default class CheckVisitDetailsController {
...
+            validationErrors.includes('APPLICATION_INVALID_PRISONER_NOT_FOUND') ||
+            validationErrors.includes('APPLICATION_INVALID_PRISON_PRISONER_MISMATCH')
+          ) {
+            return next(error)
+          }
+
+          if (validationErrors.includes('APPLICATION_INVALID_NO_VO_BALANCE')) {
+            return res.redirect(paths.BOOK_VISIT.CANNOT_BOOK)
+          }
+
```
tpmcgowan, line 66: It's likely that these checks and new error code responses will be added to the create/change application in the future. When they are, these conditions can be moved to a separate helper function to avoid duplication.

integration_tests/e2e/bookingJourney.cy.ts
tpmcgowan, line 1: No actual change to the Cypress tests - just splitting out the booking journey tests and the 'drop-out' points tests to separate files to simplify.

hutcheonb-moj (dismissed):
LGTM

server/routes/bookVisit/cannotBookController.ts
```diff
@@ -9,10 +8,6 @@ export default class CannotBookController {
       const { prisoner } = req.session.bookingJourney
       delete req.session.bookingJourney
 
-      if (prisoner.availableVos >= 1) {
```
tpmcgowan, line 12: Removing this check as the data could be stale and it will be re-fetched if user goes back to home page.

PR #134, opened 2024-07-23
VB-3853, VB-4056 Use GOV.UK One Login post_logout_redirect for user sign out

tpmcgowan:
Implement the `post_logout_redirect` GOV.UK One Login functionality (documented [here](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/managing-your-users-sessions/#log-your-user-out-of-gov-uk-one-login)) so that when a user signs out they are ultimately redirected back to `/sign-out` on our service. The previous behaviour was to end-up with a 'signed-out' page on One Login's domain.

Includes:
* refactor of `TokenStore` and a new `tokenStoreFactory` to get either an in-memory or Redis store with a customisable key prefix. This is needed because a user's `id_token` needs to be stored when they log in so that it is available for use in the log out process
* update of One Login wiremock endpoints to handle storing the `id_token` across requests
* tidy of remaining hard-coded paths

### DIFF
--- a/integration_tests/mockApis/govukOneLogin.ts
+++ b/integration_tests/mockApis/govukOneLogin.ts
@@ -4,6 +4,8 @@ import { Response } from 'superagent'
 import { createPublicKey } from 'crypto'
 import { getMatchingRequests, stubFor } from './wiremock'
 
+let idToken: string
+
 const oidcConfig = JSON.parse(
   fs.readFileSync('integration_tests/mockApis/mappings/openid-configuration.json').toString(),
 )
@@ -80,12 +82,13 @@ const createIdToken = (nonce: string) => {
   }
 
   const privateKey = fs.readFileSync('integration_tests/testKeys/server_private_key.pem')
-  const idToken = jwt.sign(payload, privateKey, { algorithm: 'ES256' })
-  return idToken
+  idToken = jwt.sign(payload, privateKey, { algorithm: 'ES256' })
 }
 
-const token = (nonce: string) =>
-  stubFor({
+const token = (nonce: string) => {
+  createIdToken(nonce)
+
+  return stubFor({
     request: {
       method: 'POST',
       url: '/govukOneLogin/token',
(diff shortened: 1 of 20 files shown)


integration_tests/mockApis/govukOneLogin.ts
```diff
@@ -4,6 +4,8 @@ import { Response } from 'superagent'
 import { createPublicKey } from 'crypto'
 import { getMatchingRequests, stubFor } from './wiremock'
 
+let idToken: string
```
tpmcgowan, line 7: Declare this here so it can be set via the sign-in stub and then used in the sign-out one.

server/data/tokenStore/tokenStoreFactory.ts
```diff
@@ -0,0 +1,18 @@
+import config from '../../config'
+import { createRedisClient } from '../redisClient'
+import InMemoryTokenStore from './inMemoryTokenStore'
+import RedisTokenStore from './redisTokenStore'
+import TokenStore from './tokenStore'
+
+let inMemoryTokenStore: InMemoryTokenStore
+
+const tokenStoreFactory = (prefix: string): TokenStore => {
```
tpmcgowan, line 9: Adding this so the in-memory vs Redis store logic doesn't need to be repeated wherever a store is created.

server/middleware/setUpGovukOneLogin.ts
```diff
@@ -38,11 +40,19 @@ export default function setUpGovukOneLogin(): Router {
...
+        const idToken = await idTokenStore.getToken(encodeURIComponent(req.user.sub))
         req.logout(err => {
           if (err) return next(err)
-          return req.session.destroy(() => res.redirect(client.endSessionUrl()))
+          return req.session.destroy(() =>
+            res.redirect(
+              client.endSessionUrl({
+                id_token_hint: idToken,
+                post_logout_redirect_uri: `${config.domain}${paths.SIGNED_OUT}`,
+              }),
```
tpmcgowan, line 53: The crux of this change: pass the token and logout redirect URI when calling OL's log out.

psoleckimoj (approved):
LGTM!

PR #138, opened 2024-07-24
VB-3615/VB-3616/VB-3617/VB-3618/ Future and Cancelled visits views

hutcheonb-moj:
Draft PR 

List Past and Cancelled visit pages

Add necessary parts to the visit details page to deal with both Cancelled and Past visits

### DIFF
--- a/server/services/visitService.test.ts
+++ b/server/services/visitService.test.ts
@@ -33,6 +33,14 @@ describe('Visit service', () => {
     const application = TestData.applicationDto()
     const visit = TestData.visitDto()
     const orchestrationVisitDto = TestData.orchestrationVisitDto()
+    const cancelledVisit = TestData.orchestrationVisitDto({
+      outcomeStatus: 'ADMINISTRATIVE_CANCELLATION',
+      visitStatus: 'CANCELLED',
+    })
+    const pastVisit = TestData.orchestrationVisitDto({
+      startTimestamp: '2023-05-30T10:00:00',
+      endTimestamp: '2023-05-30T11:30:00',
+    })
     const visitSession = TestData.availableVisitSessionDto()
 
     let bookingJourney: BookingJourney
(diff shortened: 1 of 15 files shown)


server/views/pages/bookings/visit.njk
```diff
@@ -14,6 +24,35 @@
...
+            
+        {% set cancelledBy %}
+          {% if status[0] === 'VISITOR' %}
+            {{- 'This visit was cancelled by a visitor.' -}}
+          {% elif status[0] === 'PRISONER' %}
+            {{- 'This visit was cancelled by the prisoner.' -}}
+          {% else %}
+            {{- 'This visit was cancelled by the prison.' -}}
+          {% endif %}
+        {% endset %}
```
tpmcgowan, line 39: Looking at the definition of `outcomeStatus` it's got 20 values and more than 1 would match either `visitor_` or `prisoner_`. Assuming this logic is from the staff service but we might want to check which values **actually** get set and whether this logic is still correct or whether we should be matching specific values.

server/services/visitService.test.ts
```diff
@@ -33,6 +33,14 @@ describe('Visit service', () => {
...
     const visit = TestData.visitDto()
     const orchestrationVisitDto = TestData.orchestrationVisitDto()
+    const cancelledVisit = TestData.orchestrationVisitDto({
+      outcomeStatus: 'ADMINISTRATIVE_CANCELLATION',
+      visitStatus: 'CANCELLED',
+    })
+    const pastVisit = TestData.orchestrationVisitDto({
+      startTimestamp: '2023-05-30T10:00:00',
+      endTimestamp: '2023-05-30T11:30:00',
+    })
```
tpmcgowan, line 43: This fine, but I guess not strictly necessary - i.e. it doesn't matter if the object you're testing these with is *really* a past/cancelled visit as it's not what's covered by the assertions. It just needs to be a 'Visit-shaped' object. What you're asserting below is that the given service method calls the correct data client method.

tpmcgowan: n.b. it would be different if the service method had any logic within it that looked at, say, `outcomeStatus`

tpmcgowan (approved):
LGTM!

PR #139, opened 2024-07-25
VB-4266, send username when retrieving available visit sessions

hutcheonb-moj:


integration_tests/mockApis/orchestration.ts
```diff
@@ -210,6 +212,7 @@ export default {
           prisonId: { equalTo: prisonId },
           prisonerId: { equalTo: prisonerId },
           visitors: { equalTo: visitorIds.join(',') },
+          username,
```
tpmcgowan, line 215:
```suggestion
          username: { equalTo: username },
```
I think this is probably why the tests are failing

server/data/orchestrationApiClient.ts
```diff
@@ -139,18 +139,21 @@ export default class OrchestrationApiClient {
...
     excludedApplicationReference?: string
+    username: string
   }): Promise<AvailableVisitSessionDto[]> {
     return this.restClient.get({
       path: '/visit-sessions/available',
       query: new URLSearchParams({
         prisonId,
         prisonerId,
         visitors: visitorIds.join(','),
+        username,
```
tpmcgowan, line 156:
```suggestion
    bookerReference,
  }: {
    prisonId: string
    prisonerId: string
    visitorIds: number[]
    excludedApplicationReference?: string
    bookerReference: string
  }): Promise<AvailableVisitSessionDto[]> {
    return this.restClient.get({
      path: '/visit-sessions/available',
      query: new URLSearchParams({
        prisonId,
        prisonerId,
        visitors: visitorIds.join(','),
        username: bookerReference,
```

Personal preference I guess, but it might make more sense within our application to use `bookerReference` rather than `username` when we're passing it around here and in other files  - because that's what it actually is. Then right at the point of sending to the API we map it to `username`.

tpmcgowan (approved):
LGTM!

PR #142, opened 2024-07-29
VB-4264 Remove analytics cookies when declining cookie consent

tpmcgowan:
If a user has previously accepted Google Analytics but then subsequently chooses to reject it, they are left with `_ga and _ga_[containerId]` cookies still set.

This change removes these two cookies when analytics consent is declined, either by the form on the Cookies page (so it's done server-side) or via the cookie banner (so done client-side).

### DIFF
--- a/assets/js/cookieBanner.js
+++ b/assets/js/cookieBanner.js
@@ -1,4 +1,4 @@
-(() => {
+;(() => {
   function confirmCookiesChoice(acceptAnalytics) {
     document.getElementById('cookie-banner-main').hidden = true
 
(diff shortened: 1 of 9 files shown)


assets/js/cookieBanner.js
```diff
@@ -21,10 +21,21 @@
...
+  }
+
+  function removeAnalyticsCookies() {
+    const expires = new Date(0).toUTCString()
+    const gaID = document
+      .querySelector('[data-google-analytics-id]')
+      ?.getAttribute('data-google-analytics-id')
+      .replace('G-', '')
+    document.cookie = `_ga=; expires=${expires}; path=/;`
+    document.cookie = `_ga_${gaID}=; expires=${expires}; path=/;`
```
tpmcgowan, line 38: These are the actual intended changes to this file, to remove cookies (the rest is just `prettier`)

assets/js/cookieBanner.js
```diff
@@ -21,10 +21,22 @@
...
+    banner
+      .querySelectorAll('#cookie-banner-accepted, #cookie-banner-rejected')
+      .forEach(message => (message.hidden = true))
 
-      banner.hidden = true
+    banner.hidden = true
+  }
+
+  function removeAnalyticsCookies() {
+    const domain = location.hostname === 'localhost' ? 'localhost' : 'justice.gov.uk'
```
tpmcgowan, line 32: By default, GA seems to set cookies for `justice.gov.uk` - unless it's localhost....

psoleckimoj: Is it possible to configure this in GA? I wonder if it's our account is set to cover all that domain? It just means the GA cookie is accessible from any site on that domain I think?

tpmcgowan: Yeah, that's right - seems it's a Google default to go for a fairly high-level sub-domain. Most other HMPPS services seem to be leaving the default set as far as I can tell because you accumulate a lot of `_ga_[containerId]` if you do a bit of browsing around DPS and linked services. One Login does similar on their domains too. I guess it enables tracking journeys across DPS services, for example 🤷‍♂️ ?

I'll leave it as it is for now (as you've approved it) but will raise on the ticket to confirm whether we should be setting a more specific GA cookie domain.

PR #144, opened 2024-07-31
Add integration tests for past/cancelled visits pages

hutcheonb-moj:
Add integration tests for past and cancelled visits pages


integration_tests/e2e/bookings.cy.ts
```diff
@@ -0,0 +1,139 @@
...
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignIn')
+    cy.task('stubHmppsAuthToken')
+  })
+
+  it('should show Bookings home page with future visits and navigate to view the visit details', () => {
+    cy.task('stubGetBookerReference')
+    cy.task('stubGetPrisoners', { prisoners: [prisoner] })
+    cy.signIn()
```
tpmcgowan, line 41: These three lines are now repeated in each test so could add them to the `beforeEach()` block - works fine as is though.

tpmcgowan (dismissed):
LGTM!

PR #148, opened 2024-08-09
VB-3525 / VB-3526 / View assigned visitors page

hutcheonb-moj:
Create new page to view all visitors assigned to booker account


psoleckimoj (approved):
LGTM!

tpmcgowan (changes requested):
Two issues to resolve before this can be merged:
1. It doesn't show banned visitors as specified on the ticket. API changes will be needed to support this - I'll create a Jira task
2. It assumes booker/prisoner in the session and errors if they're not present. Triggered for example by being on the page, letting session time out and then attempting refresh

tpmcgowan (approved):
LGTM

PR #157, opened 2024-09-04
VB-3815 Clear session when abandoning journey

tpmcgowan:
Add a route to clear session and redirect back to home when abandoning journey on the closed visit interruption card.

Also, refactor to use this session clearing method in other places.


hutcheonb-moj (approved):
LGTM

PR #171, opened 2024-09-23
VB-4199 Use UUIDs for Prisoner and Visitor display IDs

tpmcgowan:
In web forms, a `displayId` is used for Prisoner and Visitor objects. This is to avoid using the real database ID. 

Previously, this value was just a sequential integer. This change uses a UUID instead for these values.

(Actual code change very small and pretty much just in `server/services/bookerService.ts` - rest is updating tests...)

### DIFF
--- a/server/services/bookerService.test.ts
+++ b/server/services/bookerService.test.ts
@@ -4,6 +4,16 @@ import { createMockHmppsAuthClient, createMockOrchestrationApiClient } from '../
 
 const token = 'some token'
 
+jest.mock('uuid', () => {
+  let count = 0
+  return {
+    v4: () => {
+      count += 1
+      return `uuidv4-${count}`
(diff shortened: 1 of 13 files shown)


server/services/bookerService.test.ts
```diff
@@ -4,6 +4,16 @@ import { createMockHmppsAuthClient, createMockOrchestrationApiClient } from '../
 
 const token = 'some token'
 
+let uuidCount: number
+jest.mock('uuid', () => {
+  return {
+    v4: () => {
+      uuidCount += 1
+      return `uuidv4-${uuidCount}`
```
tpmcgowan, line 12: Return a known series of UUIDs for testing

psoleckimoj (approved):
LGTM!

PR #172, opened 2024-09-24
VB-4199 Use UUIDs for referencing visit bookings in URL parameters

tpmcgowan:
For bookings listing pages of future / past / cancelled visits, individual visits are referenced by a URL parameter. To avoid using the actual visit reference, this was previously just a sequential integer.

This change is to instead use a UUID as a URL parameter to reference a temporary cache of visits in the session.

Additionally, refactoring of controllers for booking listings and booking details to remove duplication.

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -1,6 +1,6 @@
 import { ValidationError } from 'express-validator'
 import { Booker, BookingConfirmed, BookingJourney } from '../bapv'
-import { OrchestrationVisitDto } from '../../data/orchestrationApiTypes'
+import { VisitDetails } from '../../services/visitService'
 
 export default {}
 
@@ -12,12 +12,12 @@ declare module 'express-session' {
 
     booker: Booker
 
-    bookingsFuture?: OrchestrationVisitDto[]
-    bookingsPast?: OrchestrationVisitDto[]
-    bookingsCancelled?: OrchestrationVisitDto[]
-
     bookingJourney?: BookingJourney
     bookingConfirmed?: BookingConfirmed
+
+    bookingsFuture?: VisitDetails[]
+    bookingsPast?: VisitDetails[]
+    bookingsCancelled?: VisitDetails[]
   }
 }
 
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -15,9 +15,10 @@ declare module 'express-session' {
     bookingJourney?: BookingJourney
     bookingConfirmed?: BookingConfirmed
 
-    bookingsFuture?: VisitDetails[]
-    bookingsPast?: VisitDetails[]
-    bookingsCancelled?: VisitDetails[]
+    bookings?: {
+      type: 'future' | 'past' | 'cancelled'
+      visits: VisitDetails[]
+    }
   }
 }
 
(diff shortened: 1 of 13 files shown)


server/@types/express/index.d.ts
```diff
@@ -12,12 +12,13 @@ declare module 'express-session' {
...
     booker: Booker
 
-    bookingsFuture?: OrchestrationVisitDto[]
-    bookingsPast?: OrchestrationVisitDto[]
-    bookingsCancelled?: OrchestrationVisitDto[]
-
     bookingJourney?: BookingJourney
     bookingConfirmed?: BookingConfirmed
+
+    bookings?: {
```
tpmcgowan, line 18: Refactor how temporary cache of bookings is stored in session so typed `type` can be leveraged to remove duplication in controllers.

server/services/visitService.ts
```diff
@@ -1,8 +1,12 @@
+import { v4 as uuidv4 } from 'uuid'
 import logger from '../../logger'
 import { BookingJourney } from '../@types/bapv'
 import { RestClientBuilder, OrchestrationApiClient, HmppsAuthClient } from '../data'
 import { ApplicationDto, OrchestrationVisitDto, VisitDto } from '../data/orchestrationApiTypes'
 
+export interface VisitDetails extends OrchestrationVisitDto {
```
tpmcgowan, line 7: New `VisitDetails` type which has additional `visitDisplayId` to hold UUID generated to reference it.

server/routes/bookings/bookingsController.ts
```diff
@@ -1,46 +1,40 @@
...
 import { VisitService } from '../../services'
 import getPrisonInformation from '../../constants/prisonInformation'
+import { VisitDetails } from '../../services/visitService'
 
 export default class BookingsController {
   public constructor(private readonly visitService: VisitService) {}
 
-  public viewFuture(): RequestHandler {
-    return async (req, res) => {
+  public view(type: SessionData['bookings']['type']): RequestHandler {
```
tpmcgowan, line 11: Just one method now to handle future / past / cancelled booking listings, with `switch` to get the appropriate data.

server/routes/bookings/bookingDetailsController.test.ts
```diff
@@ -2,148 +2,185 @@ import type { Express } from 'express'
 import request from 'supertest'
 import * as cheerio from 'cheerio'
 import { SessionData } from 'express-session'
+import { randomUUID } from 'crypto'
```
tpmcgowan, line 5: Discovered that you can get a UUID (v4) directly from Node's `crypto` module so no need to have `uuid` module as an NPM dependency. Will sort this elsewhere in a subsequent PR...

psoleckimoj: Hmm, could move the template repo over to use this also 🤔

server/services/visitService.ts
```diff
@@ -1,8 +1,12 @@
+import { v4 as uuidv4 } from 'uuid'
```
psoleckimoj, line 1: Could you not use the same uuid from crypto here?

server/routes/bookings/bookingDetailsController.test.ts
tpmcgowan: Yeah, I can have a look at that - can trial it on our projects to be sure it works!

psoleckimoj (approved):
LGTM! Just the one comment about uuid

server/services/visitService.ts
tpmcgowan: Yes - I think I'd written this code (and its tests) before realising. Going to sort here and other places in next PR to be consistent and get rid of `uuid` module.

PR #175, opened 2024-09-24
VB-4199 Remove 'uuid' dependency and replace with crypto.randomUUID()

tpmcgowan:
* Remove `uuid` dependency and use Node's `crypto.randomUUID()` (as advised on https://www.npmjs.com/package/uuid)

* Update Select visitors validation code to use `express-validator`'s `isUUID()` validation


psoleckimoj (approved):
LGTM!

PR #184, opened 2024-09-30
VB-4536 - Remove h2 from cancelled visits and past visits page in public service

simonnebMOJ:
Updated pages to remove the H2 that were not in the original designs.

### DIFF
--- a/assets/scss/local.scss
+++ b/assets/scss/local.scss
@@ -1,11 +1,3 @@
 .govuk-main-wrapper {
   min-height: 300px;
 }
-
-.bookings-content-block > p, .bookings-content-block > h3 {
-  margin-bottom: govuk-spacing(1);
-}
-.bookings-content-block {
-  margin-top: govuk-spacing(4);
-  margin-bottom: govuk-spacing(6);
-}
--- a/server/views/pages/bookings/cancelled.njk
+++ b/server/views/pages/bookings/cancelled.njk
@@ -13,8 +13,6 @@
 
       <p>Only bookings made using GOV.UK One Login are shown.</p>
 
-      <h2 class="govuk-heading-m">{{ pageTitle }}</h2>
-
       {% if visits | length %}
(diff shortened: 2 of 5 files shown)


tpmcgowan:
Hmm, interesting 🤔 . Looks like integration tests failing the automated accessibility checks because we've now got H1 -> H3 on a page with no H2 in between. I'll take a look!

server/views/pages/bookings/cancelled.njk
```diff
@@ -13,22 +13,17 @@
...
-              <span data-test="visit-start-time-{{ loop.index }}">{{ visit.startTimestamp | formatTimeFromDateTime }}</span>
-              to
-              <span data-test="visit-end-time-{{ loop.index }}">{{ visit.endTimestamp | formatTimeFromDateTime }}</span>
-            </p>
-            <p>
-              <a href="{{ paths.BOOKINGS.VISIT_CANCELLED }}/{{ visit.visitDisplayId }}" class="govuk-link--no-visited-state" data-test="visit-link-{{ loop.index }}">View booking details</a>
-            </p>
-          </div>
-          
+          <h2 class="govuk-heading-s govuk-!-margin-bottom-1" data-test="visit-date-{{ loop.index }}">{{ visit.startTimestamp | formatDate(dateFormats.PRETTY_DATE) }}</h2>
```
tpmcgowan, line 18: With the `h2` above removed, this now needs to be an `h2` rather than `h3` to keep correct headings order (and visually remains same size).

assets/scss/local.scss
```diff
@@ -1,11 +1,3 @@
 .govuk-main-wrapper {
   min-height: 300px;
 }
-
-.bookings-content-block > p, .bookings-content-block > h3 {
```
tpmcgowan, line 5: Rather than extend this to style `h2`s too, removing in favour of a few GOVUK margin override classes - simpler markup overall.

PR #185, opened 2024-10-04
VB-4336 Use prison contact information from API

tpmcgowan:
Remove hard-coded contact details for the two pilot prisons and retrieve prison contact details from the API so any onboarded prison can be supported.

Handle case where a prison has no visits phone number.

Tidy and update tests.


server/routes/bookings/bookingsController.ts
```diff
@@ -1,21 +1,28 @@
...
   public view(type: SessionData['bookings']['type']): RequestHandler {
     return async (req, res, next) => {
       const { booker } = req.session
 
       let visits: VisitDetails[]
+      let prison: PrisonDto
       switch (type) {
         case 'future':
           visits = await this.visitService.getFuturePublicVisits(booker.reference)
+          // making assumption here that all visits[] will be for the same prison
```
tpmcgowan, line 23: This can be addressed when the service supports more than one prisoner per booker.

PR #187, opened 2024-10-07
VB-4336 Remove duplication for change booking text

tpmcgowan:
Remove duplication for "How to change your booking" text (and fix an inconsistency between the three instances of it).


hutcheonb-moj (approved):
LGTM

PR #191, opened 2024-10-15
VB-4715 Fix encoding of JSON string cookie values (and update dependencies)

tpmcgowan:
Updates to the `cookie` module failed because it now has stricter validation on cookie value fields. Setting cookies was failing because only `JSON.stringify()` was being used.

Now uses `encodeURIComponent() / decodeURIComponent()` too in order to fix this issue.


psoleckimoj (approved):
LGTM!

PR #195, opened 2024-10-18
VB-4469, Booker cancel visit journey

hutcheonb-moj:
Add cancel journey for booker made visits

Show cancel option until the start of the visit

### DIFF
--- a/server/constants/paths.ts
+++ b/server/constants/paths.ts
@@ -27,6 +27,8 @@ const paths = {
     VISIT: '/bookings/details',
     VISIT_PAST: '/bookings/past/details',
     VISIT_CANCELLED: '/bookings/cancelled/details',
+
+    // Cancel journey
     CANCEL_VISIT: '/bookings/cancel-booking',
     CANCEL_CONFIRMATION: '/bookings/booking-cancelled',
   },
--- a/server/constants/paths.ts
+++ b/server/constants/paths.ts
@@ -27,6 +27,10 @@ const paths = {
     VISIT: '/bookings/details',
     VISIT_PAST: '/bookings/past/details',
     VISIT_CANCELLED: '/bookings/cancelled/details',
+    CANCEL: {
+      CANCEL_VISIT: '/bookings/cancel-booking',
+      CANCELLED_VISIT: '/bookings/booking-cancelled',
+    },
   },
 
   VISITORS: '/visitors',
--- a/server/constants/paths.ts
+++ b/server/constants/paths.ts
@@ -27,10 +27,8 @@ const paths = {
     VISIT: '/bookings/details',
     VISIT_PAST: '/bookings/past/details',
     VISIT_CANCELLED: '/bookings/cancelled/details',
-    CANCEL: {
-      CANCEL_VISIT: '/bookings/cancel-booking',
-      CANCELLED_VISIT: '/bookings/booking-cancelled',
-    },
+    CANCEL_VISIT: '/bookings/cancel-booking',
+    CANCEL_CONFIRMATION: '/bookings/booking-cancelled',
   },
 
   VISITORS: '/visitors',
--- a/server/routes/bookings/bookingDetailsController.ts
+++ b/server/routes/bookings/bookingDetailsController.ts
@@ -30,12 +30,17 @@ export default class BookingDetailsController {
 
       const prison = await this.prisonService.getPrison(visit.prisonId)
 
+      const nowTimestamp = new Date()
+      const visitStartTimestamp = new Date(visit.startTimestamp)
+      const showCancel = nowTimestamp < visitStartTimestamp
+
       return res.render('pages/bookings/visit', {
         booker,
         prison,
         prisoner,
         type,
         visit,
+        showCancel,
         showServiceNav: true,
       })
     }
--- a/server/routes/bookings/bookingsController.test.ts
+++ b/server/routes/bookings/bookingsController.test.ts
@@ -54,6 +54,9 @@ describe('Bookings homepage (future visits)', () => {
         expect($('[data-test="visit-link-booking-1"]').attr('href')).toBe(
           `${paths.BOOKINGS.VISIT}/${futureVisitDetails.visitDisplayId}`,
         )
+        expect($('[data-test="visit-link-cancel-1"]').attr('href')).toBe(
+          `${paths.BOOKINGS.CANCEL_VISIT}/${futureVisitDetails.visitDisplayId}`,
+        )
 
         expect($('[data-test=change-booking-heading]').length).toBeTruthy()
         expect($('[data-test="prison-name"]').text()).toBe(prison.prisonName)
@@ -122,9 +125,10 @@ describe('Past visits page', () => {
         expect($('[data-test="visit-date-1"]').text()).toBe('Thursday 30 May 2024')
         expect($('[data-test="visit-start-time-1"]').text()).toBe('10am')
         expect($('[data-test="visit-end-time-1"]').text()).toBe('11:30am')
-        expect($('[data-test="visit-link-1"]').attr('href')).toBe(
+        expect($('[data-test="visit-link-booking-1"]').attr('href')).toBe(
           `${paths.BOOKINGS.VISIT_PAST}/${pastVisitDetails.visitDisplayId}`,
         )
+        expect($('[data-test="visit-link-cancel-1"]').attr('href')).toBe(undefined)
 
         expect($('[data-test="no-visits"]').length).toBeFalsy()
 
@@ -189,9 +193,10 @@ describe('Cancelled visits page', () => {
         expect($('[data-test="visit-date-1"]').text()).toBe('Thursday 30 May 2024')
         expect($('[data-test="visit-start-time-1"]').text()).toBe('10am')
         expect($('[data-test="visit-end-time-1"]').text()).toBe('11:30am')
-        expect($('[data-test="visit-link-1"]').attr('href')).toBe(
+        expect($('[data-test="visit-link-booking-1"]').attr('href')).toBe(
           `${paths.BOOKINGS.VISIT_CANCELLED}/${cancelledVisitDetails.visitDisplayId}`,
         )
+        expect($('[data-test="visit-link-cancel-1"]').attr('href')).toBe(undefined)
 
         expect($('[data-test="no-visits"]').length).toBeFalsy()
 
--- a/server/routes/bookings/bookingsController.test.ts
+++ b/server/routes/bookings/bookingsController.test.ts
@@ -51,7 +51,7 @@ describe('Bookings homepage (future visits)', () => {
         expect($('[data-test="visit-start-time-1"]').text()).toBe('10am')
         expect($('[data-test="visit-end-time-1"]').text()).toBe('11:30am')
         expect($('[data-test="visit-reference-1"]').text()).toBe('ab-cd-ef-gh')
-        expect($('[data-test="visit-link-1"]').attr('href')).toBe(
+        expect($('[data-test="visit-link-booking-1"]').attr('href')).toBe(
           `${paths.BOOKINGS.VISIT}/${futureVisitDetails.visitDisplayId}`,
         )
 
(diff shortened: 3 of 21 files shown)


server/constants/paths.ts
```diff
@@ -27,6 +27,10 @@ const paths = {
     VISIT: '/bookings/details',
     VISIT_PAST: '/bookings/past/details',
     VISIT_CANCELLED: '/bookings/cancelled/details',
+    CANCEL: {
```
psoleckimoj, line 30: Does this need a sub object? We don't have it for `VISIT_*` ?

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,98 @@
...
+    private readonly bookerService: BookerService,
+    private readonly visitService: VisitService,
+  ) {}
+
+  public confirmCancelView(): RequestHandler {
+    return async (req, res) => {
+      const { booker, bookings } = req.session
+      const { visits } = bookings
+
+      const prisoner = booker.prisoners?.[0]
```
psoleckimoj, line 17: Probably better just doing `booker.prisoners?.length > 0` ?

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,98 @@
...
+        ? booker.prisoners?.[0]
+        : (await this.bookerService.getPrisoners(booker.reference))?.[0]
```
psoleckimoj, line 19: Also could this come back with no prisoners? So trying to access `[0]` will cause a fail?

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      const { visitDisplayId } = matchedData<{ visitDisplayId: string }>(req)
+
+      const visit = visits.find(v => v.visitDisplayId === visitDisplayId)
+
+      const nowTimestamp = new Date()
```
psoleckimoj, line 25: Maybe a util fn as you're doing this twice? Just pass the comparison date and return true/false?

server/views/pages/bookings/visit.njk
```diff
@@ -71,6 +72,16 @@
       <p class="govuk-!-margin-bottom-1" data-test="main-contact-name">{{ visit.visitContact.name }}</p>
       <p data-test="main-contact-number">{{ visit.visitContact.telephone | d("No phone number provided", true) }}</p>
 
+      {% if showCancel %}
+        {{ govukButton({
+          text: "Cancel booking",
+          classes: "govuk-!-margin-top-5 govuk-button--secondary",
+          href: paths.BOOKINGS.CANCEL.CANCEL_VISIT + '/' + visit.visitDisplayId,
```
psoleckimoj, line 79: This is why I suggested removing the upper `{ CANCEL: ...` bit, `CANCEL.CANCEL_VISIT` looks strange, either `CANCEL.VISIT` or `CANCEL_VISIT`, with the latter matching what everything else does.

server/views/pages/bookings/future.njk
```diff
@@ -25,7 +25,8 @@
             Booking reference: <span data-test="visit-reference-{{ loop.index }}">{{ visit.reference }}</span>
           </p>
           <p>
-            <a href="{{ paths.BOOKINGS.VISIT }}/{{ visit.visitDisplayId }}" class="govuk-link--no-visited-state" data-test="visit-link-{{ loop.index }}">View booking details</a>
+            <a href="{{ paths.BOOKINGS.VISIT }}/{{ visit.visitDisplayId }}" class="govuk-link--no-visited-state" data-test="visit-link-booking-{{ loop.index }}">View booking details</a>
+            <a href="{{ paths.BOOKINGS.CANCEL_VISIT }}/{{ visit.visitDisplayId }}" class="govuk-link--no-visited-state govuk-!-padding-left-3" data-test="visit-link-cancel-{{ loop.index }}">Cancel booking</a>
```
tpmcgowan, line 29: Figma design has some hidden content for accessibility that should be incorporated into these links. (Also, text in the design is now `"View details"`, not `"View booking details"`.)

server/views/pages/bookings/cancel/cancel.njk
```diff
@@ -0,0 +1,70 @@
...
+      </p>
+
+      <h2 class="govuk-heading-m">Prisoner</h2>
+
+      <p class="govuk-!-margin-bottom-1" data-test="prisoner-name">{{ prisoner.firstName | capitalize }} {{ prisoner.lastName | capitalize }}</p>
+
+      <h2 class="govuk-heading-m">Visitors</h2>
+
+      {% for visitor in visit.visitors %}
+        <p data-test="visitor-name-{{ loop.index }}">{{ visitor.firstName }} {{ visitor.lastName }}
```
tpmcgowan, line 33: Missing `</p>`

server/views/pages/bookings/cancel/cancel.njk
```diff
@@ -0,0 +1,70 @@
...
+              }
+            }
+          ],
+          errorMessage: errors | findError('cancelBooking')
+        }) }}
+          
+        {{ govukButton({
+          text: "Confirm",
+          attributes: {
+            "data-test": "continue-button"
```
tpmcgowan, line 63: Preference really, but is `confirm-button` more accurate 🤷‍♂️?

server/services/visitService.ts
```diff
@@ -90,6 +90,22 @@ export default class VisitService {
...
+    applicationReference,
+    actionedBy,
+  }: {
+    applicationReference: string
+    actionedBy: string
+  }): Promise<VisitDto> {
+    const token = await this.hmppsAuthClient.getSystemClientToken()
+    const orchestrationApiClient = this.orchestrationApiClientFactory(token)
+
+    const visit = orchestrationApiClient.cancelVisit({ applicationReference, actionedBy })
```
tpmcgowan, line 103:
```suggestion
    const visit = await orchestrationApiClient.cancelVisit({ applicationReference, actionedBy })
```

server/services/visitService.ts
```diff
@@ -90,6 +90,22 @@ export default class VisitService {
...
+
+    logger.info(`Visit '${applicationReference}' has been cancelled by booker '${actionedBy}`)
+    return visit
```
tpmcgowan, line 106: The returned `visit` isn't actually used in the controller. So this method could just return `void` (same for the corresponding  data layer method)

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -30,12 +30,17 @@ export default class BookingDetailsController {
...
+      const visitStartTimestamp = new Date(visit.startTimestamp)
+      const showCancel = nowTimestamp < visitStartTimestamp
+
       return res.render('pages/bookings/visit', {
         booker,
         prison,
         prisoner,
         type,
         visit,
+        showCancel,
```
tpmcgowan, line 43: Could add to `bookingDetailsController.test.ts` to assert that cancel button **not** present on past/cancelled details view and **is** present on future details view.

server/routes/bookings/bookingsController.test.ts
```diff
@@ -51,7 +51,7 @@ describe('Bookings homepage (future visits)', () => {
         expect($('[data-test="visit-start-time-1"]').text()).toBe('10am')
         expect($('[data-test="visit-end-time-1"]').text()).toBe('11:30am')
         expect($('[data-test="visit-reference-1"]').text()).toBe('ab-cd-ef-gh')
-        expect($('[data-test="visit-link-1"]').attr('href')).toBe(
+        expect($('[data-test="visit-link-booking-1"]').attr('href')).toBe(
```
tpmcgowan, line 54: Should probably also assert that the new `Cancel visit` link is present on listings for future bookings, but not on past/cancelled listings.

server/routes/bookings/cancelVisitController.ts
tpmcgowan: Yes, potentially. Have spoken to @hutcheonb-moj about this. We've got the same block of code in a few places now. It's all to do with needing a prisoner's name to display. I've arranged to get a change to the API done (which is in progress) so we get prisoner name returned with the visit data itself so no longer need to fish it out from the booker record in the session.

So, OK for now I think as better, less brittle solution coming immanently!

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,112 @@
...
+      await this.visitService.cancelVisit({
+        applicationReference: visit.reference,
+        actionedBy: booker.reference,
+      })
+
+      return res.redirect(paths.BOOKINGS.CANCEL_CONFIRMATION)
+    }
+  }
+
+  public visitCancelled(): RequestHandler {
```
tpmcgowan, line 76: Should probably have a test for this route (it'll be a very simple one!)

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+
+  public submit(): RequestHandler {
+    return async (req, res, next) => {
+      const { booker, bookings } = req.session
+      const { visits } = bookings
+      const { cancelBooking } = req.body
+
+      const { visitDisplayId } = matchedData<{ visitDisplayId: string }>(req)
```
tpmcgowan, line 47:
```suggestion
      const { cancelBooking, visitDisplayId } = matchedData<{ cancelBooking: 'yes' | 'no', visitDisplayId: string }>(req)
```

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      const errors = validationResult(req)
+      if (!errors.isEmpty()) {
+        req.flash('errors', errors.array())
+
+        if (!visitDisplayId) {
+          return res.redirect(paths.BOOKINGS.HOME)
+        }
+
+        req.flash('formValues', matchedData(req, { onlyValidData: false }))
```
tpmcgowan, line 57: Might not need this line 🤔. No free text fields or similar to pre-populate: so either it was the expected `yes|no` or it's something unexpected we don't want to hang on to?

tpmcgowan: Also, not actually reading `formValues` from flash in the GET method - so it should go. Otherwise it'll be left 'orphaned' in flash and potentially picked up elsewhere.

server/routes/bookings/cancelVisitController.ts
```diff
@@ -0,0 +1,112 @@
...
+        return res.redirect(`${paths.BOOKINGS.CANCEL_VISIT}/${visitDisplayId}`)
+      }
+
+      const visit = visits.find(v => v.visitDisplayId === visitDisplayId)
```
tpmcgowan, line 61: Perhaps move this down a few lines: no point in looking up the visit if they've said 'no' to cancelling.

tpmcgowan (commented):
Looking good. I've added a few comments. Also:
* probably worth creating a feature branch (e.g. `feature/VB-4268_booker-cancels-visit`) to merge this into so the story's other sub-tasks can be merged into that too.
* needs an integration test

server/routes/bookings/cancelVisitController.ts
hutcheonb-moj: is this not replacing the functionality introduced by the validation?

tpmcgowan: No, it's typing it. So, if you check the type of `cancelBooking` having got it straight from `req.body` it'll be something like `any`. In reality, the actual *value* could be anything - because invalid data will be present on `req.body`. (By using `matchedData` with an expected type passed in, `cancelBooking` will have the type `'yes '| 'no'` (see docs [here](https://express-validator.github.io/docs/api/matched-data/#typescript-usage)).

We can be sure it is actually this (and not optional) because it wouldn't appear in `matchedData` if it was empty or an unexpected value.

tpmcgowan (approved):
LGTM 👍 . Needs an integration test for this new journey; but approving so you can get this on dev for testing and do integration test separately if you wish.

PR #206, opened 2024-11-01
Hide cancel visit button when visit is already cancelled

hutcheonb-moj:
Hide cancel visit button when visit is already cancelled
(Visits in the past already don't display the button)

Fix logic on the howToChangeBooking import, to deal with 'visit details page' and the 'future visits list page'
Both of these don't have the 'bookingConfirmed' object, so this PR deals with the different available visit references, and outputs them IF that line should be displayed

### DIFF
--- a/server/views/partials/howToChangeBooking.njk
+++ b/server/views/partials/howToChangeBooking.njk
@@ -1,4 +1,10 @@
 <h2 class="govuk-heading-m" data-test="change-booking-heading">How to change your booking</h2>
+{% set visitRef = '' %}
+{% if bookingConfirmed.visitReference %}
+  {% set visitRef = bookingConfirmed.visitReference %}
+{% elif visit.reference %}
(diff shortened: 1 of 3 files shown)


server/views/partials/howToChangeBooking.njk
```diff
@@ -1,4 +1,10 @@
 <h2 class="govuk-heading-m" data-test="change-booking-heading">How to change your booking</h2>
+{% set visitRef = '' %}
+{% if bookingConfirmed.visitReference %}
+  {% set visitRef = bookingConfirmed.visitReference %}
+{% elif visit.reference %}
+  {% set visitRef = visit.reference %}
+{% endif %}
```
tpmcgowan, line 7:
```suggestion
{% set visitRef = bookingConfirmed.visitReference else visit.reference  %}
```
Might work?

tpmcgowan (approved):
LGTM!

PR #209, opened 2024-11-05
VB-4775 'Booking cancelled' page: handle optional text message info

tpmcgowan:
* Booking cancelled page: optionally show message about receiving text notification depending on whether phone number provided
* Small refactor (split existing controller - primarily to keep tests simpler)
* Fix pre-existing bug that could lead to a crash if `bookings` not present in session


server/@types/bapv.d.ts
```diff
@@ -49,3 +49,7 @@ export type BookingConfirmed = {
   visitReference: string
   hasPhoneNumber: boolean
 }
+
+export type BookingCancelled = {
```
tpmcgowan, line 53: Using an object here rather than just a simple boolean for the phone number because we will be expanding to have email and/or phone.

server/routes/bookings/bookingDetailsController.ts
```diff
@@ -26,6 +25,7 @@ export default class BookingDetailsController {
 
       const { visitDisplayId } = matchedData<{ visitDisplayId: string }>(req)
 
+      const { visits } = bookings
```
tpmcgowan, line 28: Unrelated, but spotted pre-existing bug here (and also present in cancel controllers). `bookings` may not be present (e.g. reloading from an expired session) so need to check for validation errors **before** trying to restructure `visits` from it.

server/routes/bookings/cancel/cancelController.ts
tpmcgowan, line 1: Moved these controllers into a `cancel` sub-folder - to match how templates are in `views/bookings/cancel`

server/routes/bookings/cancel/cancelController.ts
```diff
@@ -62,22 +60,18 @@ export default class CancelVisitController {
...
 
       await this.visitService.cancelVisit({
         applicationReference: visit.reference,
         actionedBy: booker.reference,
       })
 
-      return res.redirect(paths.BOOKINGS.CANCEL_CONFIRMATION)
-    }
-  }
+      req.session.bookingCancelled = { hasPhoneNumber: !!visit.visitContact.telephone }
```
tpmcgowan, line 72: Main part of the change: set session data when cancelling so we know if a phone number is available

server/routes/bookings/cancel/cancelConfirmedController.ts
tpmcgowan, line 1: I've split this route for the 'booking cancelled' page to a new controller. Main reason is that the tests can then be simpler because this one doesn't need other dependencies mocking, etc.

hutcheonb-moj (approved):
LGTM

PR #210, opened 2024-11-07
VB-4524 Add service navigation when user not signed in

simonnebMOJ:
We need to implement the new “service navigation” component alongside the new simplified GOV.UK Header component in the public service.

[Service navigation – GOV.UK Design System (design-system.service.gov.uk)](https://design-system.service.gov.uk/components/service-navigation/)

[GOV.UK header – GOV.UK Design System (design-system.service.gov.uk)](https://design-system.service.gov.uk/components/header/)

We also need to move elements currently in the Header component into the new “Service navigation” component. 

GDS will retire the ‘old’ Header in a future break changes version of the Design System.

---
There are some pages in the new public service that we need to update - particularly if they are being viewed by users who are NOT signed in to the service. These are:

* cookies page
* terms and conditions page
* accessibility page
* privacy policy page

### DIFF
--- a/server/routes/unauthenticatedRoutes.test.ts
+++ b/server/routes/unauthenticatedRoutes.test.ts
@@ -64,7 +64,7 @@ describe('Accessibility statement', () => {
 
         expect($('header.govuk-header').length).toBe(1)
         expect($('header .one-login-header').length).toBe(0)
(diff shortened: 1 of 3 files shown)


tpmcgowan (approved):
LGTM!

PR #217, opened 2024-11-14
Removed reference to cancelling

simonnebMOJ:
Cancellations can now be done in the service so no need to refer bookers to other journeys to cancel a visit.


tpmcgowan (approved):
LGTM!

PR #219, opened 2024-11-18
VB-4839 Update visit booked page

tpmcgowan:
* Add `libphonenumber-js` as a dependency in order to be able to determine if numbers are mobile or not
* Update Visit booked confirmation page to
  * handle different messages depending on contact details provided (email/phone/both/neither)
  * distinguish between mobile and other numbers so "text message will be sent" displayed only if a mobile number was supplied
* Update integration tests for new Contact details page

### DIFF
--- a/integration_tests/e2e/bookingJourney.cy.ts
+++ b/integration_tests/e2e/bookingJourney.cy.ts
@@ -143,10 +143,10 @@ context.skip('Booking journey', () => {
     const mainContactPage = Page.verifyOnPage(MainContactPage)
     mainContactPage.selectVisitorByName('Adult One')
     mainContactPage.checkHasPhoneNumber()
-    mainContactPage.enterPhoneNumber('01234 567 890')
+    mainContactPage.enterPhoneNumber('07700 900 000')
     cy.task('stubChangeVisitApplication', {
       ...application,
-      visitContact: { name: 'Adult One', telephone: '01234 567 890' },
+      visitContact: { name: 'Adult One', telephone: '07700 900 000' },
       visitors: [
         { nomisPersonId: 1000, visitContact: true },
         { nomisPersonId: 3000, visitContact: false },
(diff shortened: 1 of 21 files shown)


integration_tests/pages/page.ts
```diff
@@ -4,8 +4,8 @@ import logAccessibilityViolations from '../support/logAccessibilityViolations'
 export type PageElement = Cypress.Chainable<JQuery>
 
 export default abstract class Page {
-  static verifyOnPage<T>(constructor: new () => T): T {
-    return new constructor()
+  static verifyOnPage<T>(constructor: new (title?: string) => T, title?: string): T {
```
tpmcgowan, line 7: Updating this so we can optionally pass a title for integration test Page object constructor, for cases where page titles vary.

integration_tests/e2e/bookingJourney.cy.ts
```diff
@@ -142,18 +143,24 @@ context.skip('Booking journey', () => {
...
+    // Contact details
+    const contactDetailsPage = Page.verifyOnPage(ContactDetailsPage, 'Adult One')
+    contactDetailsPage.checkGetUpdatesByEmail()
+    contactDetailsPage.enterEmail('adult.one@example.com')
+    contactDetailsPage.checkGetUpdatesByPhone()
+    contactDetailsPage.enterPhoneNumber('07712 000 000')
     cy.task('stubChangeVisitApplication', {
       ...application,
-      visitContact: { name: 'Adult One', telephone: '01234 567 890' },
+      visitContact: { name: 'Adult One', telephone: '07712 000 000', email: 'adult.one@example.com' },
```
tpmcgowan, line 156: Updating the dummy number used in various test files to be a mobile one, so it enables, e.g. messages about texts being sent that are now only shown if the number type is mobile.

(Originally tried to use a number in Ofcom's `07700 900000 to 900999` 'dummy' unallocated range but these failed `libphonenumber-js`'s validation. ChatGTP suggested `07712 000 000` - which it says has a valid mobile prefix but invalid subscriber portion 😅 !)

psoleckimoj (approved):
lgtm

integration_tests/e2e/bookingJourney.cy.ts
tpmcgowan: yeah, tried those ones (they're on the Ofcom site - I found them linked to from the GOVUK Notify API docs) but they failed **because** they're unallocated I think. This one that ChatGTP suggested seems to work because it's a valid operator prefix but apparently the `000000` bit would never be allocated 🤷‍♂️

PR #223, opened 2024-11-22
VB-4907 Improve handling of authentication errors

tpmcgowan:
Calls to `passport.authenticate()` in the `/auth/callback` route can fail if, for example, multiple attempts are made with the same parameters (`code, state`), or, if expired ones are used. Evidence in logs of this happening to users, possibly from, e.g. returning to stale browser tabs.

This change adds a callback to handle the response (whether successful or not) for `passport.authenticate()`. This means that errors in the authentication flow can be logged and handled appropriately by redirecting the user to either the sign-in or authentication error page. Previously errors would just trigger the default error handler resulting in an `HTTP 500` response with a default server error page.

(Similar approach to that used in a few other projects, e.g. https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/34bc9a3e5b84011267e9d09c74d21ddd9be501b3/server/middleware/setUpAuthentication.ts#L32)

### DIFF
--- a/integration_tests/mockApis/govukOneLogin.ts
+++ b/integration_tests/mockApis/govukOneLogin.ts
@@ -3,6 +3,7 @@ import jwt, { JwtPayload } from 'jsonwebtoken'
 import { Response } from 'superagent'
 import { createPublicKey } from 'crypto'
 import { getMatchingRequests, stubFor } from './wiremock'
+import paths from '../../server/constants/paths'
 
 let idToken: string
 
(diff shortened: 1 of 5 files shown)


integration_tests/mockApis/govukOneLogin.ts
```diff
@@ -41,8 +42,7 @@ const redirect = () =>
         scope: { equalTo: 'openid email phone' },
         client_id: { equalTo: 'clientId' },
         state: { matches: '.*' },
-        redirect_uri: { equalTo: 'http://localhost:3007/auth/callback' },
-        nonce: { matches: '.*' },
```
tpmcgowan, line 45: Remove match on `nonce` as it is not always passed (and other tests check that when it is it is used correctly).

psoleckimoj (approved):
LGTM!

psoleckimoj:
Might be worth a back port into the template?

tpmcgowan:
> Might be worth a back port into the template?

I'll have a look to see if we're getting similar issues in staff service (ie with HMPPS Auth) as that's closer to the template...

PR #230, opened 2024-12-03
VB-4362 Refactor cannot book page to handle different scenarios

tpmcgowan:
Refactor the 'A visit cannot be booked page' and associated controller/middleware to work with an extensible set of `CannotBookReason` values - initially just `NO_VO_BALANCE` supported.


hutcheonb-moj (approved):
LGTM

PR #231, opened 2024-12-03
VB-4758, use prisoner details from visit

hutcheonb-moj:
Use prisoner details from visit instead of booker


tpmcgowan:
I've ended up having to use some of this code in #229 so have moved this back to draft so we can see if there's anything left to bring over once that's done.

tpmcgowan:
This was all fine but closing as effectively the same changes have gone into #229

PR #232, opened 2024-12-03
VB-4362 Redirect to Cannot book page if prisoner has been transferred/released

tpmcgowan:
When starting a booking journey, call the new validate booker/prisoner endpoint and if this fails (because a prisoner has been transferred or released) redirect to the drop-out page.

Also redirect to the same drop-out page if a prisoner transfer/release is caught at the end of a booking journey (via the `APPLICATION_INVALID_PRISON_PRISONER_MISMATCH` code thrown when booking).


hutcheonb-moj (approved):
LGTM

PR #234, opened 2024-12-05
VB-4362 Incorporate API changes and use 'registered' prison details

tpmcgowan:
* Updates for changes in format of `HTTP 422` error response
* Prisoner now has details of **registered** as well as current prison. Show registered prison details on 'cannot book' page


server/@types/orchestration-api.d.ts
tpmcgowan, line 1: this file is auto-generated from Swagger

dhirajshettymoj (approved):
LGTM

PR #236, opened 2024-12-06
VB-4909, VB-4910 Show info banner if prisoner transferred or released

tpmcgowan:
On these pages:
* future visits listing
* future visit details

show an information banner if the prisoner has been released or transferred from the prison they are registered to in the service.

### DIFF
--- a/server/routes/bookings/bookingDetailsController.ts
+++ b/server/routes/bookings/bookingDetailsController.ts
@@ -1,23 +1,31 @@
 import type { RequestHandler } from 'express'
 import { Meta, ValidationChain, matchedData, param, validationResult } from 'express-validator'
 import { SessionData } from 'express-session'
-import { PrisonService } from '../../services'
+import { BookerService, PrisonService } from '../../services'
 import paths from '../../constants/paths'
 
 export default class BookingDetailsController {
-  public constructor(private readonly prisonService: PrisonService) {}
+  public constructor(
+    private readonly bookerService: BookerService,
+    private readonly prisonService: PrisonService,
+  ) {}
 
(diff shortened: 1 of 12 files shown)


server/routes/bookings/bookingDetailsController.ts
```diff
@@ -1,31 +1,39 @@
...
+  public constructor(
+    private readonly bookerService: BookerService,
+    private readonly prisonService: PrisonService,
+  ) {}
 
   public view(type: SessionData['bookings']['type']): RequestHandler {
     return async (req, res) => {
-      const { bookings } = req.session
+      const { booker, bookings } = req.session
+      const prisoner = booker.prisoners[0]
```
psoleckimoj, line 16: Is there a change this could be empty 🤔 No worries if not.

psoleckimoj: I mean not be an array?

psoleckimoj (approved):
LGTM!

server/routes/bookings/bookingDetailsController.ts
tpmcgowan: Good question - had a look and it should always be (at least an empty) array because `populateCurrentBooker` middleware runs on all these routes and sets `booker.prisoners` [here](https://github.com/ministryofjustice/hmpps-book-a-prison-visit-ui/blob/51a3817b46d872faab2f9664478ed9158958483b/server/middleware/populateCurrentBooker.ts#L22).

(Also double-checked my understanding of JS and `[0]` does in fact return `undefined` instead of an error for an empty array - and that's handled further down 👍)

PR #244, opened 2024-12-17
VB-4380 Service start page

tpmcgowan:
* Unauthenticated users initially see a new service start page when arriving at the root `/` URL. The 'Start now' button takes them through the normal sign-in flow.
* Authenticated users go straight to the services 'home' page (now actually moved to `/home`)

### DIFF
--- a/server/routes/bookVisit/visitBookedController.ts
+++ b/server/routes/bookVisit/visitBookedController.ts
@@ -9,7 +9,7 @@ export default class VisitBookedController {
       res.render('pages/bookVisit/visitBooked', {
         bookingConfirmed,
         prison: bookingConfirmed.prison,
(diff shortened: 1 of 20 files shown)


server/routes/bookVisit/visitBookedController.ts
```diff
@@ -9,7 +9,7 @@ export default class VisitBookedController {
       res.render('pages/bookVisit/visitBooked', {
         bookingConfirmed,
         prison: bookingConfirmed.prison,
-        showServiceNav: true,
+        showOLServiceNav: true,
```
tpmcgowan, line 12: Renaming this to distinguish between the **GOVUK One Login** service navigation and the standard **GOVUK service navigation** 😵‍💫

PR #251, opened 2025-01-07
VB-4390 Use MoJ Interruption card component

tpmcgowan:
Remove existing interruption card code and use new MoJ component for this.

(Plus a temporary fix for a minor styling issue with the MoJ component.)


hutcheonb-moj (approved):
LGTM

PR #273, opened 2025-03-03
VB-5197 Do not check VO balance if prisoner on REMAND

tpmcgowan:
Prisoners on remand do not have a VO balance, so ignore this when starting booking journey.


hutcheonb-moj (approved):
LGTM

PR #283, opened 2025-03-25
VB-5116 Get dynamic list of supported prisons

tpmcgowan:
Use new endpoint that returns list of supported prison IDs and names for the service. Avoids hard-coding two existing prisons and will be needed for 'Add a prisoner' journey.


hutcheonb-moj (approved):
LGTM

PR #284, opened 2025-03-26
VB-5116 Add a prisoner: select prison page

tpmcgowan:
* Remove auto redirect to add prisoner journey if booker has no prisoner: instead keep on service 'home' page with an 'Add prisoner' button
* Add Prisoner location page with choice of supported prisons (`GET` route only so far)

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -7,6 +7,12 @@ export type Booker = {
   prisoners: Prisoner[] // prisoners this booker can book for
 }
 
+// data that is built up during an add prisoner journey
+export type AddPrisonerJourney = {
+  supportedPrisonIds: string[]
+  selectedPrisonId?: string
+}
+
 // data that is built up during a booking journey
 export type BookingJourney = {
   // selected prisoner for this visit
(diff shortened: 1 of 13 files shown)


server/@types/bapv.d.ts
```diff
@@ -7,6 +7,12 @@ export type Booker = {
   prisoners: Prisoner[] // prisoners this booker can book for
 }
 
+// data that is built up during an add prisoner journey
+export type AddPrisonerJourney = {
+  supportedPrisonIds: string[]
+  selectedPrisonId?: string
```
tpmcgowan, line 13: Storing all `supportedPrisonIds` so they can be used in request validation in the `POST` route - and obviously the `selectedPrisonId` once validated (not implemented yet).

server/utils/nunjucksSetup.ts
```diff
@@ -23,6 +23,7 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
   app.locals.paths = paths
   app.locals.dateFormats = DateFormats
   app.locals.googleAnalyticsId = config.analytics.googleAnalyticsId
+  app.locals.features = config.features
```
tpmcgowan, line 26: This makes feature flags available in Nunjucks (e.g. you can do `{% if features.myFlag %}...`)

hutcheonb-moj (approved):
lgtm

PR #285, opened 2025-03-27
Improve prisoner name handling (e.g. double-barrelled names)

tpmcgowan:


server/utils/nunjucksSetup.ts
```diff
@@ -92,6 +100,10 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
     }
   })
 
+  njkEnv.addFilter('prisonerNameFirstLast', prisoner => {
```
tpmcgowan, line 103: I've made this filter specific to **prisoner** names as visitor names don't need to be processed (they don't come through all upper-case).

PR #286, opened 2025-03-27
VB-5116 Add prisoner - select location POST route and validation

tpmcgowan:
* Handle selection of prisoner location, validate and store in session
* Enable add prisoner feature in staging


hutcheonb-moj (approved):
lgtm

PR #290, opened 2025-04-01
VB-5117 Add prisoner: enter prisoner details page

tpmcgowan:
* Add API support for registering a prisoner for a booker
* 'Prisoner details' page and form validations
* Redirects to success/fail pages (not yet built - follow-on tickets)

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -11,6 +11,7 @@ export type Booker = {
 export type AddPrisonerJourney = {
   supportedPrisonIds: string[]
   selectedPrisonId?: string
+  prisonerDetails?: Record<string, string>
 }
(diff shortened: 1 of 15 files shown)


server/@types/bapv.d.ts
```diff
@@ -11,6 +11,7 @@ export type Booker = {
 export type AddPrisonerJourney = {
   supportedPrisonIds: string[]
   selectedPrisonId?: string
+  prisonerDetails?: Record<string, string>
```
tpmcgowan, line 14: This is to save the prisoner details a user has entered in case their request fails and they want to go back again and re-try after amending.

hutcheonb-moj (approved):
lgtm

PR #292, opened 2025-04-01
VB-5389 Refresh booker's prisoners when loading home page

tpmcgowan:
* Load booker's prisoners each time home page is loaded to ensure it is up-to-date - i.e. includes newly-added prisoners
* Exclude auto-loading of booker's prisoners in `populateCurrentBooker()` middleware for requests to `/home` to avoid duplicate API calls


hutcheonb-moj (approved):
LGTM

PR #293, opened 2025-04-02
VB-5117 Add prisoner: date of birth validation fix

tpmcgowan:
* Ensure day/month padded with leading zero if < 10
* Fix to ensure prisoner details are retained in session if going back to change prisoner location

### DIFF
--- a/server/routes/addPrisoner/prisonerDetailsController.test.ts
+++ b/server/routes/addPrisoner/prisonerDetailsController.test.ts
@@ -151,8 +151,8 @@ describe('Prisoner details', () => {
     const prisonerDetails = {
       firstName: 'John',
       lastName: 'Smith',
-      day: '02',
-      month: '04',
+      day: '2',
+      month: '4',
       year: '1975',
       prisonNumber: 'A1234BC',
     } as const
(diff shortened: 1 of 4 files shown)


server/routes/addPrisoner/prisonerDetailsController.test.ts
```diff
@@ -151,8 +151,8 @@ describe('Prisoner details', () => {
     const prisonerDetails = {
       firstName: 'John',
       lastName: 'Smith',
-      day: '02',
-      month: '04',
+      day: '2',
+      month: '4',
```
tpmcgowan, line 155: Worked fine in the tests - because my test data assumed dates entered with leading zeros 🤦‍♂️!

server/routes/addPrisoner/prisonerDetailsController.ts
```diff
@@ -73,32 +88,34 @@ export default class PrisonerDetailsController {
...
-        return true
-      }),
+      // 'prisonerDob' not an actual form field but used to store combined result and any validation errors
+      body('prisonerDob')
+        .customSanitizer((_value, { req }) => {
+          const day = Number.parseInt(req.body.day, 10).toString().padStart(2, '0')
+          const month = Number.parseInt(req.body.month, 10).toString().padStart(2, '0')
+          const year = Number.parseInt(req.body.year, 10).toString()
+          return `${year}-${month}-${day}`
+        })
```
tpmcgowan, line 98: making this a sanitiser function means that the combined `YYY-MM-DD` value is stored as a 'field' available in the request handler

PR #294, opened 2025-04-04
Minor text tweaks

simonnebMOJ:
Spotted a couple of small text changes.

### DIFF
--- a/server/routes/addPrisoner/prisonerDetailsController.test.ts
+++ b/server/routes/addPrisoner/prisonerDetailsController.test.ts
@@ -278,7 +278,7 @@ describe('Prisoner details', () => {
             type: 'field',
             location: 'body',
             path: 'prisonerDob',
-            value: '2025-04-03',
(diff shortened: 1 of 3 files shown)


server/routes/addPrisoner/prisonerDetailsController.test.ts
```diff
@@ -272,22 +272,23 @@ describe('Prisoner details', () => {
           })
       })
 
+      const tomorrow = addDays(new Date(), 1)
```
psoleckimoj, line 275: Might be simpler just to set the year to 2066 or something like that as I think it's just checking for a DOB in the future and we're not going to reach 2066...well, hopefully 😂

psoleckimoj: Yours is technically better as it'll never fail!

psoleckimoj (approved):
LGTM!

PR #321, opened 2025-05-22
VB-5444, add additional support character count component

hutcheonb-moj:
Switch to using character count component in public service

### DIFF
--- a/server/views/pages/bookVisit/additionalSupport.njk
+++ b/server/views/pages/bookVisit/additionalSupport.njk
@@ -26,6 +26,7 @@
             id: "additionalSupport",
             name: "additionalSupport",
             maxlength: 200,
+            threshold: '75',
             type: "text",
             label: {
               text: "Provide details of the request",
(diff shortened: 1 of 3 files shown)


server/views/pages/bookVisit/additionalSupport.njk
```diff
@@ -22,12 +22,17 @@
...
         <input type="hidden" name="_csrf" value="{{ csrfToken }}">
         {% set supportDetailsHtml %}
-          {{ govukInput({
+          {{ govukCharacterCount({
             id: "additionalSupport",
             name: "additionalSupport",
+            maxlength: 200,
             type: "text",
-            classes: "govuk-!-width-one-half",
+            classes: "govuk-!-width-two-thirds",
```
tpmcgowan, line 30: Do we need to add any `classes` here? Worth trying without as the parent column is already two-thirds...

server/views/pages/bookVisit/additionalSupport.njk
```diff
@@ -22,12 +22,17 @@
...
             label: {
+              text: "Provide details",
+              classes: "govuk-visually-hidden"
+            },
+            hint: {
```
tpmcgowan, line 35: Text on the ticket is "Provide details about the request"

tpmcgowan (changes requested):
A few things to tweak before this can go in

PR #348, opened 2025-07-03
VB-5699 Handle 'sessionForReview' flag and pass when booking

tpmcgowan:
**Only** if the `FEATURE_VISIT_REQUEST` flag is enabled (currently just dev & staging):
* Update to use available sessions V2 endpoint
* `sessionForReview` flag for selected visiting session stored in booking journey
* `isRequestBooking` flag passed in Dto when calling `/book` endpoint at end of journey

### DIFF
--- a/server/data/orchestrationApiClient.ts
+++ b/server/data/orchestrationApiClient.ts
@@ -210,7 +210,7 @@ export default class OrchestrationApiClient {
     bookerReference: string
   }): Promise<AvailableVisitSessionDto[]> {
     return this.restClient.get({
-      path: '/visit-sessions/available',
+      path: '/visit-sessions/available/v2',
       query: new URLSearchParams({
         prisonId,
         prisonerId,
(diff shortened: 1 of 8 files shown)


hutcheonb-moj (dismissed):
LGTM

server/data/orchestrationApiClient.ts
```diff
@@ -210,7 +210,7 @@ export default class OrchestrationApiClient {
     bookerReference: string
   }): Promise<AvailableVisitSessionDto[]> {
     return this.restClient.get({
-      path: '/visit-sessions/available',
+      path: config.features.visitRequest ? '/visit-sessions/available/v2' : '/visit-sessions/available',
```
tpmcgowan, line 213: This is to only use the new v2 sessions API if the feature flag is enabled (i.e. in dev & staging)

dhirajshettymoj (approved):
LGTM

PR #349, opened 2025-07-04
VB-5752 Add PVB Urls to config

tpmcgowan:
Add new config for `PVB_URL` which will be used for redirecting users to PVB for prisons that are not onboarded to the new service.


hutcheonb-moj (approved):
lgtm

PR #353, opened 2025-07-10
VB-5724 Add data caching and prisons service methods

tpmcgowan:
* Add simple cache for JSON data, with Redis and in-memory versions. (Based on similar existing TokenStore implementation)
* Add service methods for getting all prison names and identifying if a prison is supported in the service (to be used on forthcoming select a prison page)

N.B. none of this code is used yet: it's to support upcoming features.


dhirajshettymoj (approved):
LGTM - 1 small comment around a TODO comment.

PR #354, opened 2025-07-11
VB-5674 Refactor routes for unauthenticated users

tpmcgowan:
* Restructure/refactor of routes to give better structure (in preparation for adding select a prison feature)
* Parameterise tests for static content pages to remove duplication


server/routes/testutils/appSetup.ts
```diff
@@ -60,7 +59,7 @@ function appSetup(
 
   nunjucksSetup(app, testAppInfo)
   app.use((req, res, next) => {
-    if (!sessionData.booker && populateBooker) {
+    if (!sessionData.booker) {
```
tpmcgowan, line 62: This change is just to a testing helper: `populateBooker` was a way to mock a user being logged in to One Login but without a booker reference. Now that we auto-create booker references the first time a user logs in this scenario is redundant. So removing from helper and some tests.

server/routes/unauthenticatedRoutes.test.ts
```diff
@@ -55,198 +55,3 @@ describe('Service start page', () => {
     return request(app).get(paths.START).expect(302).expect('location', paths.HOME)
   })
 })
-
-describe('Accessibility statement', () => {
```
tpmcgowan, line 59: These tests have been simplified and moved to `staticPages.test.ts`

PR #355, opened 2025-07-11
VB-5674, VB-5760 Select prison and selected prison page, with autocomplete

tpmcgowan:
* Add pages for 'Select prison' journey
  * for supported prisons a user is sent to a prison info page and then logs in to the service
  * for unsupported prisons the user is redirected to PVB
* Integration tests for these journeys
* Added helper methods for mocking feature flags in tests

N.B. This journey is only enabled in dev and staging currently because of the `FEATURE_VISIT_REQUEST` flag.

### DIFF
--- a/server/middleware/setUpWebSecurity.ts
+++ b/server/middleware/setUpWebSecurity.ts
@@ -41,6 +41,7 @@ export default function setUpWebSecurity(): Router {
             "'self'",
             config.apis.govukOneLogin.url,
             config.apis.govukOneLogin.url.replace('oidc', 'signin'),
+            config.pvbUrl,
           ],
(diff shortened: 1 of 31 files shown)


server/middleware/setUpWebSecurity.ts
```diff
@@ -41,6 +41,7 @@ export default function setUpWebSecurity(): Router {
             "'self'",
             config.apis.govukOneLogin.url,
             config.apis.govukOneLogin.url.replace('oidc', 'signin'),
+            config.pvbUrl,
```
tpmcgowan, line 44: This is needed here because if PVB url is not in the CSP, browsers won't follow the redirect.

hutcheonb-moj (dismissed):
lgtm

PR #374, opened 2025-08-07
Fixing incorrect page width

simonnebMOJ:
During testing, we noticed that the "Visit/request booking details" page was incorrectly set to full width. This PR fixes it to 2/3 width in line with the other pages.


adaviesMOJ (approved):

psoleckimoj (approved):
LGTM!

PR #379, opened 2025-08-18
VB-5841, display visit request tags

hutcheonb-moj:
Display tags to highlight to visitors that these visits are request

### DIFF
--- a/server/views/pages/bookings/future.njk
+++ b/server/views/pages/bookings/future.njk
@@ -3,7 +3,7 @@
 
 {% set activeLink = 'bookings-link' %}
 {% set pageTitle = "Bookings" %}
-{% set referenceType = "Booking reference" %}
+{% set referenceType = "Request reference" if visit.visitSubStatus == "REQUESTED" else "Booking reference" %}
 
 
 {% block content %}


assets/scss/local.scss
```diff
@@ -24,6 +24,10 @@
   color: govuk-colour("red");
 }
 
+.tag-margin-bottom {
+  margin-bottom: 10px;
+}
```
tpmcgowan, line 29: I would just use `govuk-!-margin-bottom-2` instead in the one place where you're using `tag-margin-bottom` as it's basically doing the same thing but without the need to define an extra class.

server/views/pages/bookings/visit.njk
```diff
@@ -4,19 +4,47 @@
...
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
 
-      {# Messages #}
-      {% for message in messages %}
-        {{ mojAlert(message) }}
-      {% endfor %}
+      {% if isRequest or isRejected %}
+        {% set pageTitle = "Visit request details" %}
+        {% set referenceType = "Request reference" %}
```
tpmcgowan, line 18: isn't this (partially) duplicating what's above?

server/views/pages/bookings/visit.njk
```diff
@@ -4,19 +4,47 @@
...
+        {% elif isRejected %}
+          {{ mojAlert({
+            variant: "information",
+            dismissible: false,
+            text: "This request was rejected by " + prison.prisonName + "."
+          }) }}
+        {% endif %}
+
+      {% else %}
+        {# Messages only shown on non-requested type visits #}
```
tpmcgowan, line 37: A lot of nested logic in the template here. `messages` are supposed to be a set of `MoJ Alert(s)` that can be shown on the page. So might be simpler and easier to test to add them as appropriate in the controller (either directly or via a utility function).

server/views/pages/bookings/future.njk
```diff
@@ -1,7 +1,10 @@
 {% extends "partials/layout.njk" %}
+{%- from "govuk/components/tag/macro.njk" import govukTag -%}
 
 {% set activeLink = 'bookings-link' %}
 {% set pageTitle = "Bookings" %}
+{% set referenceType = "Booking reference" %}
```
tpmcgowan, line 6:
```suggestion
{% set referenceType = "Request reference" if visit.visitSubStatus == "REQUESTED" else "Booking reference" %}
```

If you handle the two cases here you don't need to override `referenceType` in line 28.

PR #384, opened 2025-08-22
VB-3571 Simplify and refactor handling of visitor bans

tpmcgowan:
If a visitor has multiple overlapping `BAN` restrictions, the API processes these so only a single `BAN` would be received with the furthest (or indefinite) expiry date. This means code handling visitor `BAN`s can be removed/simplified.

Also:
* simplify some types
* switch to building GOVUK Table rows in utility function to simplify template code

### DIFF
--- a/server/services/bookerService.ts
+++ b/server/services/bookerService.ts
@@ -1,5 +1,6 @@
 import { randomUUID } from 'crypto'
 import { TooManyRequests } from 'http-errors'
+import { differenceInDays } from 'date-fns'
 import logger from '../../logger'
 import { HmppsAuthClient, OrchestrationApiClient, RestClientBuilder } from '../data'
 import {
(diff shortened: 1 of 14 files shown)


server/services/bookerService.ts
```diff
@@ -29,11 +28,14 @@ export type Prisoner = {
   convictedStatus?: ConvictedStatus
 }
 
-export interface Visitor extends VisitorInfoDto {
+export type Visitor = {
```
tpmcgowan, line 31: Fully define `Visitor` here instead of extending the API's type because we flatten the contents of `visitorRestrictions: []` to `banned` and `banExpiryDate` to make them easier to access.

server/views/pages/bookVisit/selectVisitors.njk
```diff
@@ -48,10 +48,10 @@
             {%- endset %}
 
             {% set visitorHintHtml -%}
-              <p class='banned-hint govuk-!-margin-bottom-0'>
+              <p class='secondary-text govuk-!-margin-bottom-0'>
```
psoleckimoj, line 51: Not specifically just this but shouldn't we be using double-quotes in HTML attributes?

psoleckimoj (dismissed):
LGTM!

server/views/pages/bookVisit/selectVisitors.njk
tpmcgowan: Fixed!

psoleckimoj (approved):
LGTM!

PR #385, opened 2025-08-22
VB-4876 Add visitor ban details to Choose visit time page

tpmcgowan:
On the 'Choose visit time' (calendar) page, add inset text component with details of visitor ban(s) that are affecting the available sessions shown.

If multiple visitors with bans are selected only the visitor(s) with the ban expiring last are shown.


psoleckimoj (approved):
LGTM!

PR #389, opened 2025-08-29
VB-5888 Fix implementation of error summaries/components

tpmcgowan:
Fix handling of links in GOV.UK Error Summary components to link correctly to relevant input fields, to match the [guidance](https://design-system.service.gov.uk/components/error-summary/).

Previously, error summary links were linking to a field's error message component rather than the field itself.

(The accessibility audit identified this specifically in relation to a date input, but it turned out to be a more general problem.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,8 +133,6 @@
     "@types/passport": "^1.0.17",
     "@types/superagent": "^8.1.9",
     "@types/supertest": "^6.0.3",
-    "@typescript-eslint/eslint-plugin": "^8.41.0",
-    "@typescript-eslint/parser": "^8.41.0",
     "audit-ci": "^7.1.0",
     "axe-core": "^4.10.3",
     "cheerio": "^1.1.2",
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -11,7 +11,14 @@ export type Booker = {
 export type AddPrisonerJourney = {
   supportedPrisons: PrisonRegisterPrisonDto[]
   selectedPrison?: PrisonRegisterPrisonDto
-  prisonerDetails?: Record<string, string>
+  prisonerDetails?: {
+    firstName: string
+    lastName: string
+    'prisonerDob-day': string
+    'prisonerDob-month': string
+    'prisonerDob-year': string
+    prisonNumber: string
+  }
   result?: boolean
 }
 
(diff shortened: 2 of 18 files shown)


server/@types/bapv.d.ts
```diff
@@ -11,7 +11,14 @@ export type Booker = {
 export type AddPrisonerJourney = {
   supportedPrisons: PrisonRegisterPrisonDto[]
   selectedPrison?: PrisonRegisterPrisonDto
-  prisonerDetails?: Record<string, string>
+  prisonerDetails?: {
```
tpmcgowan, line 14: Improving typing of this, which is used to store data during the 'Add a prisoner' journey. 

Not keen on the property names that need quoting but difficult to avoid as the `govukDateInput` names fields in this way.

server/routes/addPrisoner/prisonerDetailsController.ts
```diff
@@ -87,35 +87,37 @@ export default class PrisonerDetailsController {
...
+      body(['prisonerDob-day', 'prisonerDob-month', 'prisonerDob-year']).trim(),
+      // 'prisonerDob' not an actual form field but used to store combined result
+      body('prisonerDob').customSanitizer((_value, { req }) => {
+        const day = Number.parseInt(req.body?.['prisonerDob-day'], 10).toString().padStart(2, '0')
+        const month = Number.parseInt(req.body?.['prisonerDob-month'], 10).toString().padStart(2, '0')
+        const year = Number.parseInt(req.body?.['prisonerDob-year'], 10).toString()
+        return `${year}-${month}-${day}`
+      }),
+      // set any validation errors against the 'day' field so ErrorSummary links to this
+      body('prisonerDob-day').custom((_value, { req }) => {
```
tpmcgowan, line 100: Key change here: return the combined `YYY-MM-DD` field as `prisonerDob` but accumulate any errors against just the 'day' input so this is what the error summary links to.

package.json
```diff
@@ -133,8 +133,6 @@
     "@types/passport": "^1.0.17",
     "@types/superagent": "^8.1.9",
     "@types/supertest": "^6.0.3",
-    "@typescript-eslint/eslint-plugin": "^8.41.0",
-    "@typescript-eslint/parser": "^8.41.0",
```
tpmcgowan, line 137: Unrelated - just removing some unnecessary dependencies.

PR #540, opened 2026-03-31
VB-3748 Extract text to locale files - add a visitor journey

tpmcgowan:


server/views/pages/addPrisoner/prisonerDetails.njk
```diff
@@ -1,6 +1,6 @@
 {% extends "partials/layout.njk" %}
 {% from "govuk/components/input/macro.njk" import govukInput %}
-{% from "components/dateOfBirthInput.njk" import dateOfBirthInput %}
+{% from "components/dateOfBirthInput.njk" import dateOfBirthInput with context %}
```
tpmcgowan, line 3: The `with context` option when importing a macro lets it have access to scope so it can use the `t()` function internally.

hutcheonb-moj (approved):
lgtm

PR #555, opened 2026-04-14
Upgrade to TypeScript v6

tpmcgowan:


tsconfig.json
```diff
@@ -7,12 +7,12 @@
     "sourceMap": true,
     "noEmit": false,
     "allowJs": false,
-    "strict": true,
```
tpmcgowan, line 10: Can tidy some options because `strict: true` is new default.

psoleckimoj (approved):
lgtm!

PR #558, opened 2026-04-17
VB-6579 Handle English/Welsh prison names - use new helper everywhere

tpmcgowan:
Follow on from #557.

* Make sure all prison name rendering goes through `getPrisonName()` to ensure correct localisation
* Tidy and remove unused endpoints
* Some small refactoring to simplify, e.g. session data where prison IDs can now be stored instead of full prison data


server/services/prisonService.test.ts
```diff
@@ -127,4 +115,29 @@ describe('Prison service', () => {
       expect(orchestrationApiClient.getSupportedPrisonIds).toHaveBeenCalled()
     })
   })
+
+  describe('getSupportedPrisonIds', () => {
```
tpmcgowan, line 119: Adding a test; this is an existing method on the service but I've changed it fro `private` to `public`.

hutcheonb-moj (approved):
LGTM

PR #560, opened 2026-04-20
Lower prison name cache ttl to 1 hour

tpmcgowan:
Lower cache time so any changes come though more quickly.


dhirajshettymoj (approved):
LGTM

PR #561, opened 2026-04-22
VB-3748 Move remaining English strings to locale files

tpmcgowan:
Get remaining text into locale files; mainly alert messages, banners, etc.

### DIFF
--- a/server/middleware/setUpI18n.ts
+++ b/server/middleware/setUpI18n.ts
@@ -11,6 +11,9 @@ export default function setUpI18n({ production }: { production: boolean }): Rout
   const router = express.Router()
 
   if (!i18next.isInitialized) {
+    const cookieExpirationDate = new Date()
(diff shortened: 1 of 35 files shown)


server/middleware/setUpI18n.ts
```diff
@@ -11,6 +11,9 @@ export default function setUpI18n({ production }: { production: boolean }): Rout
   const router = express.Router()
 
   if (!i18next.isInitialized) {
+    const cookieExpirationDate = new Date()
+    cookieExpirationDate.setFullYear(cookieExpirationDate.getFullYear() + 1) // 1 year
```
tpmcgowan, line 15: Set `lng` cookie expiration to 1 year explicitly rather than relying on `i18next` default.

hutcheonb-moj (approved):
lgtm

PR #566, opened 2026-04-23
Remove generic telephone opening times from public service

simonnebMOJ:
Removes generic telephone opening times from public service pages


tpmcgowan (approved):
LGTM!

PR #626, opened 2026-07-02
VB-6811 Add JSON/.xlsx converter for translations

tpmcgowan:
The application uses JSON files for English / Welsh locales. These are not easy for our translators to work with. Their preference is to have these in Excel format, with JSON keys flattened and one translation per row.

This change adds some (Copilot CLI generated) tooling to convert between the application's internal JSON format. The scripts are documented in the READMEs and are intended to be run only on developers' local installations. There is no CI automation, for example.

### DIFF
--- a/bin/tsconfig.json
+++ b/bin/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "extends": "@tsconfig/node24/tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "allowImportingTsExtensions": true,
+    "resolveJsonModule": true,
+    "types": ["node"]
+  },
+  "include": ["**/*.ts"]
+}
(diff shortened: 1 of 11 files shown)


bin/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "@tsconfig/node24/tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "allowImportingTsExtensions": true,
```
tpmcgowan, line 5: Custom TypeScript config for these `.ts` scripts in `bin/` because they are run natively with Node rather than being compiled with `tsc`.

package.json
```diff
@@ -22,10 +22,12 @@
     "start-feature:dev": "concurrently -k -p \"[{name}]\" -n \"Views,Locales,TypeScript,Node,Sass\" -c \"yellow.bold,magenta.bold,cyan.bold,green.bold,blue.bold\" \"npm run watch-views\" \"npm run watch-locales\" \"npm run watch-ts\" \"npm run watch-node-feature\" \"npm run watch-sass\"",
     "lint": "eslint . --cache --max-warnings 0",
     "lint-fix": "eslint . --cache --max-warnings 0 --fix",
-    "typecheck": "tsc && tsc -p integration_tests",
+    "typecheck": "tsc && tsc -p integration_tests && tsc -p bin",
     "test": "jest",
     "test:ci": "jest --runInBand",
     "security_audit": "npx audit-ci --config audit-ci.json",
+    "i18n:export-xlsx": "node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON bin/i18nExportXlsx.ts",
+    "i18n:import-xlsx": "node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON bin/i18nImportXlsx.ts",
```
tpmcgowan, line 30: Warnings disabled here as the new scripts are just standalone and don't meet the requirements of the rest of the TypeScript code in the app.

hutcheonb-moj (approved):
lgtm

PR #628, opened 2026-07-06
VB-6779 Refactor data, token and rate-limit caches

tpmcgowan:
The service uses several Redis-backed caches: authentication tokens, API data cache and rate limit counts. (These can fallback to in-memory equivalents for tests or local development.)

This change separates these out more explicitly and leaves the code in `./server/data/tokenStore` as it was originally in the TypeScript template app.

This means that subsequently upgrading to the HMPPS shared Auth and Rest client libraries should be simpler.


hutcheonb-moj (approved):
lgtm

PR #633, opened 2026-07-17
VB-6844 Add script (from Copilot) to compare en/cy locale files

tpmcgowan:
Script (generated by Copilot) to compare the set of English and Welsh translations. It will report on:
* any JSON files missing from either locale
* any translation keys missing from either locale

Usage: `npm run i18n:locale-compare`. 

E.g.
```
$ npm run i18n:locale-compare            

Files only in 'en': newFeature.json

bookVisit.json:
  Keys only in 'en':
    bookVisit:cannotBook.newButton, Button for a new thing
    bookVisit:cannotBook.newLabel, Label for a new thing
```

### DIFF
--- a/bin/i18nLocaleCompare.ts
+++ b/bin/i18nLocaleCompare.ts
@@ -64,11 +64,15 @@ function main() {
       console.log(`\n${fileName}:`)
       if (enOnlyKeys.length) {
         console.log(`  Keys only in 'en':`)
-        enOnlyKeys.forEach(key => console.log(`    ${namespace}:${key}, ${enEntries.get(key)}`))
+        enOnlyKeys.forEach(key =>
+          console.log(`    [\x1b[31m${namespace}:${key}\x1b[0m]: \x1b[34m${enEntries.get(key)}\x1b[0m`),
+        )
       }
       if (cyOnlyKeys.length) {
         console.log(`  Keys only in 'cy':`)
-        cyOnlyKeys.forEach(key => console.log(`    ${namespace}:${key}, ${cyEntries.get(key)}`))
+        cyOnlyKeys.forEach(key =>
+          console.log(`    [\x1b[31m${namespace}:${key}\x1b[0m]: \x1b[34m${cyEntries.get(key)}\x1b[0m`),
+        )
       }
     }
   })
--- a/bin/i18nLocaleCompare.ts
+++ b/bin/i18nLocaleCompare.ts
@@ -1,0 +1,80 @@
+/* eslint-disable no-console */
+/* eslint-disable import/extensions */
+
+/**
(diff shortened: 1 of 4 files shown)


bin/i18nLocaleCompare.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const enOnlyKeys = [...enEntries.keys()].filter(key => !cyEntries.has(key)).sort()
+    const cyOnlyKeys = [...cyEntries.keys()].filter(key => !enEntries.has(key)).sort()
+
+    if (enOnlyKeys.length || cyOnlyKeys.length) {
+      problemsFound = true
+      console.log(`\n${fileName}:`)
+      if (enOnlyKeys.length) {
+        console.log(`  Keys only in 'en':`)
+        enOnlyKeys.forEach(key => console.log(`    ${namespace}:${key}, ${enEntries.get(key)}`))
```
sorrelingabout:
```suggestion
        enOnlyKeys.forEach(key => console.log(`    [\x1b[31m${namespace}:${key}\x1b[0m]: \x1b[34m${enEntries.get(key)}\x1b[0m`))
```

Feel free to disagree, but I feel this may be more digestible than using a comma to delimit key:value pairs in the output

```
    bookVisit:cannotBook.newButton, Button for a new thing
    [bookVisit:cannotBook.newButton]: Button for a new thing
```

The `\x1b[31m` gubbins colours the output like this when executed with Node:

Image

Should work in most shell environments, as far as I know. At any rate, I think brackets would help for readability and copy-pasting elsewhere. Colours are just for being fancy reasons :)

What do you think?

bin/i18nLocaleCompare.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+      if (cyOnlyKeys.length) {
+        console.log(`  Keys only in 'cy':`)
+        cyOnlyKeys.forEach(key => console.log(`    ${namespace}:${key}, ${cyEntries.get(key)}`))
```
sorrelingabout:
```suggestion
        cyOnlyKeys.forEach(key => console.log(`    [\x1b[31m${namespace}:${key}\x1b[0m]: \x1b[34m${cyEntries.get(key)}\x1b[0m`))
```

sorrelingabout (dismissed):
Small suggestion, but otherwise looks good to me!

bin/i18nLocaleCompare.ts
sorrelingabout: _[Note: other colours are available](https://gist.github.com/JBlond/2fea43a3049b38287e5e9cefc87b2124)_ 🌈

tpmcgowan: Yeah, why not! Suggestions applied 👍

sorrelingabout (approved):
LGTM 😄

PR #634, opened 2026-07-17
Use UUID type for visitDisplayId

tpmcgowan:
Improve typing of `visiDisplayId` to use the `UUID` type - and for consistency with how this is done for visitor/prisoner display IDs.


sorrelingabout (approved):
LGTM

PR #638, opened 2026-07-20
VB-6847 Fix language-toggle on add visitor / prisoner confirmation pages

tpmcgowan:
The English / Welsh language toggle works by reloading the current page URL with the requested language query parameter.

Previously the confirmation pages (success/failure) for adding a prisoner or visitor could be rendered once-only because the session data for these journeys was destroyed in order to prevent a user going back and re-submitting. This meant that language toggling was broken on these pages.

This change defines two new session data items to store the results of the add prisoner/visitor journeys so that the main journey data can be deleted at the end but data for the confirmation page can persist.

### DIFF
--- a/server/routes/addPrisoner/prisonerDetailsController.ts
+++ b/server/routes/addPrisoner/prisonerDetailsController.ts
@@ -62,9 +62,13 @@ export default class PrisonerDetailsController {
         prisonId: addPrisonerJourney.selectedPrisonId,
       })
 
-      addPrisonerJourney.result = result
+      req.session.addPrisonerJourneyResult = {
+        selectedPrisonId: addPrisonerJourney.selectedPrisonId,
+        result,
+      }
 
       if (result) {
+        delete req.session.addPrisonerJourney
         return res.redirect(paths.ADD_PRISONER.SUCCESS)
       }
 
--- a/server/routes/addVisitor/checkVisitorDetailsController.ts
+++ b/server/routes/addVisitor/checkVisitorDetailsController.ts
@@ -10,7 +10,7 @@ export default class CheckVisitorDetailsController {
       const { addVisitorJourney } = req.session
 
       if (!addVisitorJourney) {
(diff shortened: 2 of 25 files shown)


server/routes/addVisitor/checkVisitorDetailsController.ts
```diff
@@ -10,7 +10,7 @@ export default class CheckVisitorDetailsController {
       const { addVisitorJourney } = req.session
 
       if (!addVisitorJourney) {
-        return res.redirect(paths.ADD_VISITOR.DETAILS)
+        return res.redirect(paths.VISITORS)
```
tpmcgowan, line 13: Unrelated but spotted this was redirecting to the wrong page in the absence of correct session data.

server/views/pages/visitors/visitors.njk
```diff
@@ -15,7 +15,7 @@
       <h2 class="govuk-heading-m" data-test="prisoner-visitors">
         {{- t("visitors:visitors.visitorsOf", {
           name: prisoner | prisonerNameFirstLast
-        }) | safe -}}
+        }) -}}
```
tpmcgowan, line 18: Unrelated - spotted that this was incorrectly being filtered through `safe`.

server/routes/addPrisoner/prisonerDetailsController.ts
```diff
@@ -62,9 +62,13 @@ export default class PrisonerDetailsController {
...
 
-      addPrisonerJourney.result = result
+      req.session.addPrisonerJourneyResult = {
+        selectedPrisonId: addPrisonerJourney.selectedPrisonId,
+        result,
+      }
 
       if (result) {
+        delete req.session.addPrisonerJourney
         return res.redirect(paths.ADD_PRISONER.SUCCESS)
```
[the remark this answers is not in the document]
tpmcgowan: Fixed

adaviesMOJ (approved):

psoleckimoj (approved):

PR #640, opened 2026-07-22
VB-6786, VB-6788: Use rest-client, auth-client-modules

psoleckimoj:


server/data/orchestrationApiClient.test.ts
```diff
@@ -354,7 +359,7 @@ describe('orchestrationApiClient', () => {
     })
   })
 
-  describe('getPrisoners', () => {
+  describe('validatePrisoner', () => {
     it('should retrieve prisoners associated with a booker', async () => {
```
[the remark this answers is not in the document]
psoleckimoj: Fixed separately

server/data/index.ts
```diff
@@ -1,36 +1,33 @@
...
 
 const applicationInfo = applicationInfoSupplier()
-
-type RestClientBuilder<T> = (token: string) => T
-
 const redisClient = config.redis.enabled ? createRedisClient() : null
+const hmppsAuthClient = new AuthenticationClient(
+  config.apis.hmppsAuth,
+  logger,
+  config.redis.enabled ? new RedisTokenStore(createRedisClient()) : new InMemoryTokenStore(),
```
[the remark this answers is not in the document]
psoleckimoj: Separately used suggestion

server/data/orchestrationApiClient.test.ts
```diff
@@ -369,8 +374,8 @@ describe('orchestrationApiClient', () => {
     })
   })
 
-  describe('validatePrisoner', () => {
-    it('should call validate endpoint for given prisoner and booker reference', async () => {
+  describe('getPrisoners', () => {
```
psoleckimoj, line 377: These were incorrectly named in the tests and the wrong way round

server/data/orchestrationApiClient.ts
```diff
@@ -22,15 +24,12 @@ import {
...
 
 export type SessionRestriction = AvailableVisitSessionDto['sessionRestriction']
 
-export default class OrchestrationApiClient {
-  private restClient: RestClient
-
-  constructor(token: string) {
-    this.restClient = new RestClient('orchestrationApiClient', config.apis.orchestration as ApiConfig, token)
+export default class OrchestrationApiClient extends RestClient {
+  constructor(authenticationClient: AuthenticationClient | undefined) {
```
psoleckimoj, line 31: For some reason this was required, in the internal vsip app, passing `null` in the tests worked, but not here.

tpmcgowan: It's because I moved this repo over to TypeScript `strict` mode 😀 ! So it's now picker about `null`, `undefined` etc. I think someone recently added more of this kind of `| undefined` stuff to the template so it supports strict mode but I don't think it's turned on yet.

I'd quite like to move admin/staff to strict mode but haven't got round to it yet...

tpmcgowan: Thinking about this more, you might be able to fix this in the mock rather than in the actual client code here. So basically, remove the `| undefined` here and in `./data/testutils/mocks.ts` put:

```
export const createMockOrchestrationApiClient = () => new OrchestrationApiClient({} as AuthenticationClient) as jest.Mocked<OrchestrationApiClient>
```

server/services/visitService.ts
```diff
@@ -100,12 +91,9 @@ export default class VisitService {
     visitors: Visitor[]
     language: string
   }): Promise<VisitDto> {
-    const token = await this.hmppsAuthClient.getSystemClientToken()
```
tpmcgowan, line 103: Much nicer with all these duplicated sections removed!

tpmcgowan (approved):
Looks great! Much cleaner in the services now. Minor suggestion to move a an `| undefined` out of code and fix instead in mock.

server/data/orchestrationApiClient.ts
psoleckimoj: Irritatingly you still get:

```
Argument of type 'import("/hmpps-book-a-prison-visit-ui/node_modules/@ministryofjustice/hmpps-rest-client/dist/index").AuthenticationClient' is not assignable to parameter of type 'import("/hmpps-book-a-prison-visit-ui/node_modules/@ministryofjustice/hmpps-auth-clients/dist/index").AuthenticationClient'.
  Type 'AuthenticationClient' is missing the following properties from type 'AuthenticationClient': config, logger, tokenStore, name, and 18 more.ts(2345)
```

tpmcgowan: That's weird. It worked for me locally but perhaps I hadn't updated modules or something 🤔 . Need to add import `AuthenticationClient` too into the list just above. Perhaps I had `{} as unknown as AuthenticationClient`...

